PR #16, opened 2017-08-23
Drop support for Ember < 2.12.

rwjblue:
It is definitely possible to make these work if someone ultimately needs the support, but lets remove it for now.


rwjblue:
Updated to just drop 2.4 / 2.8 CI runs.

PR #17, opened 2017-08-23
Move `bower` to devDependencies.

rwjblue:
It definitely should not be a `dependency`, but I wonder if we actually need it listed in deps or devDeps at all?  I don't spot any direct `bower` invocations, and `ember-try` already installs its own copy as needed...

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,12 +18,12 @@
     "test": "ember try:each"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


hjdivad:
@rwjblue can't we just drop it completely?

hjdivad:
superseded by #24

Issue #18, opened 2017-08-23
Remove dependency on ember-data?


rwjblue:
It seems likely that we want to remove `ember-data` as a direct dependency of this addon, and instead use the host project's version, right?

- [ ] Move `ember-data` to `devDependencies`
- [ ] Add `ember-data` to `peerDependencies` 

ember-cli does not yet have any special handling of `peerDependencies`, but `npm` and `yarn` both provide nice warnings when they are not satisfied...

stefanpenner:
>  Move ember-data to devDependencies

This is my vote. With an assertion if ember-data is at the right version is not present (either from the build side, or runtime, likely build side)

PR #20, opened 2017-08-23
Move ember-data to peerdep.

hjdivad:
This partially addresses #18 but we should add a build-time assertion to make sure we're monkeypatching something we're good with, while working on making this use public API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,9 @@
   },
   "dependencies": {
     "bower": "^1.8.0",
-    "ember-cli-babel": "^6.0.0-beta.9",
+    "ember-cli-babel": "^6.0.0-beta.9"
+  },
+  "peerDependencies": {
     "ember-data": "~2.14.10"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


addon/model.js
```diff
@@ -391,6 +391,11 @@ export default class MegamorphicModel extends Ember.Object {
 
   // TODO: drop change events for unretrieved properties
   setUnknownProperty(key, value) {
+    if (key === OWNER_KEY) {
+      // 2.12 support; later versions avoid this call entirely
```
stefanpenner: is this code intended to be part of this PR?

hjdivad: @stefanpenner nope and it won't show up as part of the PR after #19 lands.  If #19 changes i'll rebase this to drop that commit.

Ah i'll just do so anyway but it means travis will for sure fail on lts 2.12

hjdivad:
@rwjblue you're right of course

rwjblue:
I restarted one of the CI jobs (all passed for the PR run, but one of them failed for the branch push run).

hjdivad:
@rwjblue yeah i had restarted that job earlier as well but it failed.  Unsure what's going on; will have to look into it once there's some timedown with a wc-test or something.

rwjblue:
I cleared all Travis caches, and restarted both branch push and PR builds.

stefanpenner:
restarting the single failed task

PR #21, opened 2017-08-23
Add missing query cache test

hjdivad:
Not a new test, but update existing test to check against this particular failure scenario

### DIFF
--- a/addon/query-cache.js
+++ b/addon/query-cache.js
@@ -48,11 +48,9 @@ export default class QueryCache {
       });
     }
 
-    // TODO: if(loadPromise)
     if (reload || cachedValue === undefined) {
       return loadPromise;
     } else {
-      // TODO: test resolve cachedValue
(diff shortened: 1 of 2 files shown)


addon/query-cache.js
```diff
@@ -48,11 +48,9 @@ export default class QueryCache {
       });
     }
 
-    // TODO: if(loadPromise)
     if (reload || cachedValue === undefined) {
```
hjdivad: Eh, I did want to simplify this conditional but it's not as simple as my earlier `TODO` implied because with both `reload` and `backgroundReload` you will have a `loadPromise` and a `cachedValue` and the thing to resolve the returned promise with will depend on which flag is set.

tl;dr I'd be happy to simplify this branching but I don't see an obvious way to do so.

stefanpenner: :shipit:

PR #22, opened 2017-08-23
Updates to README

hjdivad:
This is a first pass at docs but should be a reasonable starting point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ schema from the payload and API-specific conventions.
 
 For example, if your API returns responses like the following:
 
-```json
+```js
 {
   "data": {
     "id": "isbn:9780439708180",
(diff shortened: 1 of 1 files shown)


hjdivad:
No need for extremely detailed scrutiny but definitely happy to get feedback (on all of `README` not just the diff in this pr)

README.md
```diff
@@ -208,36 +216,349 @@ normalization to JSON API.
...
+
+Because models (or `RecordArray`s of models are cached) the cache can be emptied
+automatically when the models are unloaded.  In the case of `RecordArray`s of
+models, the entire cache entry is evicted if *any* of the member models is
+unloaded.
+
 
 ## Schema
 
+You have to register a schema to tell ember-m3 what types it should be enabled
```
stefanpenner: this should include, what is a schema, and when it is appropriate to use one, and how they differ from models.

README.md
```diff
@@ -1,4 +1,4 @@
-# ember-m3 [![Build Status](https://secure.travis-ci.org/hjdivad/ember-m3.svg?branch=master)](http://travis-ci.org/hjdivad/ember-m3)
+# Ember-m3 [![Build Status](https://secure.travis-ci.org/hjdivad/ember-m3.svg?branch=master)](http://travis-ci.org/hjdivad/ember-m3)
```
hjdivad: @rwjblue I could go either way.  I take it you're in favour of always lowercase (even in title / start of sentence)?

rwjblue: I think the only one dislike is what is here ATM (`Ember-m3`). 

IMHO, if we are referring to "the package" we should always use the package's actual name on NPM (which is `ember-m3`).  I personally would be happy to _only_ use `ember-m3`, but could also see where folks might prefer to use `Ember M3` when discussing things in prose.

Feel free to completely disregard this entire comment though 😝

hjdivad: Nope I'm convinced; I'll switch everything to `ember-m3` and consider `Ember m3` or `Ember M3` documentation bugs.

README.md
```diff
@@ -208,36 +216,349 @@ normalization to JSON API.
...
+}
+```
+
+
+### Other Computed Properties
+
+More involved computed properites can be converted to either utility functions
+(if used within JavaScript) or helper functions (if used in templates).
+
+For properties used in both templates and elsehwere (eg components) a convenient
```
jsturgis: elsehwere -> elsewhere

hjdivad: thanks @jsturgis !

PR #27, opened 2017-10-31
Fix `peekRecord`, `peekAll`

hjdivad:
Depends on unreleased ember-data 2.17.0-beta.2 although the relevant
change can be monkeypatched if needed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,10 +7,6 @@ module.exports = {
   options: {
     babel: {
       loose: true,
-    }
-  },
-
-  isDevelopingAddon() {
-    return true;
+    },
   },
 };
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -7,10 +7,9 @@ module.exports = {
   options: {
     babel: {
       loose: true,
-    }
-  },
-
-  isDevelopingAddon() {
-    return true;
+      plugins: [
+        ['ember-modules-api-polyfill', { blacklist: { '@ember/debug': ['assert', 'deprecate', 'warn']} }],
```
rwjblue: You should not need to add this manually.  It is automatically added for you when using ember-cli-babel@6.6.2 or higher.

PR #29, opened 2017-11-07
ember-try use yarn

hjdivad:
because it works


hjdivad:
https://github.com/stefanpenner/hash-for-dep/pull/30   demonstrates what's needed to work around the state `npm install` leaves `node_modules` in

PR #30, opened 2017-11-09
Drop use of bower (mostly) in ember-try

hjdivad:
Leave it only for ember-canary, although obviously it would be better to
drop it there too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ cache:
 env:
   # we recommend new addons test the current and previous LTS
   # as well as latest stable release (bonus points to beta/canary)
+  - EMBER_TRY_SCENARIO=ember-lts-2.16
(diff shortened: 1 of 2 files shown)


hjdivad:
@rwjblue there's no way to get out of bower for ember-canary right?

stefanpenner:
@hjdivad https://emberjs.com/builds/canary

stefanpenner:
@hjdivad it seems like we could easily download the artifacts from CI as part of a postinstall step or something...

rwjblue:
> @rwjblue there's no way to get out of bower for ember-canary right?

Yes, you can! As of https://github.com/emberjs/ember.js/pull/15789 you can add `"ember-source": "git:emberjs/ember.js"` to `package.json` with npm@5 or yarn.

PR #33, opened 2017-11-16
Final experiments

dnachev:
- Implement a crude merge strategy. For simplicity and performance, the strategy will produce special changed keys format, where nested keys are represented as arrays, where the first entry is the name of the key in the parent object and the rest are the keys changed in the nested object.
  The `_didReceiveNestedProperties` will skip on detecting changes if the keys have been passed.
- Remove the `@` base model as it is no longer needed. This required adjusting the payloads a little, because I think they no longer require `projectionTypes` in meta (it doesn't bring any value). The most we may need there is a flag to not mark the model as loaded.
- Remove `pushInternalModel` hook which caused two tests to be skipped (`findRecord` and `peekRecord` one). The removal of the hook also required the switch to use `computeBaseModelName()` from the schema and `recordForId()`, which will always create a base record and it solves the problem of Ember Data creating an empty projection record for the purpose of fetching it, before we have received the base record.
- Skip tests, which rely on fetching updates to a projection. The serializer is supposed to always convert these to updates to the base model and the projection. This means the test for updates to base model are already covering the above case. Only thing, which these tests should catch are changes in the state of the projection record.
- Note, because of the merge strategy, other tests are failing. I'm going to look at them immediately to make sure they are no new bugs introduced there.

### DIFF
--- a/tests/unit/projection-test.js
+++ b/tests/unit/projection-test.js
@@ -49,6 +49,11 @@ module('unit/projection', function(hooks) {
         return /^com\.example\.bookstore\./i.test(modelName) || modelName.startsWith('@');
       },
 
+      computeBaseModelName(projectionModelName) {
+        let modelSchema = this.models[projectionModelName];
+        return modelSchema && modelSchema.projectedType;
+      },
+
       computeAttributeReference(key, value, modelName) {
         if (/^isbn:/.test(value)) {
           return {
@@ -142,7 +147,7 @@ module('unit/projection', function(hooks) {
     });
   });
 
-  test(`store.peekRecord() will only return a projection or base-record if it has been fetched`, function(assert) {
+  todo(`store.peekRecord() will only return a projection or base-record if it has been fetched`, function(assert) {
     assert.expect(4);
 
     const UNFETCHED_PROJECTION_ID = 'isbn:9780439708180';
@@ -202,7 +207,7 @@ module('unit/projection', function(hooks) {
     assert.equal(record, undefined, 'The unfetched base-record with a fetched projection is unfound by peekRecord()');
   });
 
-  test(`store.findRecord() will only fetch a projection or base-model if it has not been fetched previously`, function(assert) {
+  todo(`store.findRecord() will only fetch a projection or base-model if it has not been fetched previously`, function(assert) {
     assert.expect(12);
 
     const UNFETCHED_PROJECTION_ID = 'isbn:9780439708180';
--- a/tests/unit/projection-test.js
+++ b/tests/unit/projection-test.js
@@ -187,7 +187,7 @@ module('unit/projection', function(hooks) {
               title: `Mr. Popper's Penguins`
             },
             meta: {
-              projectionTypes: [BOOK_EXCERPT_PROJECTION_CLASS_PATH]
+              partial: true,
             }
           },
         ]
@@ -278,7 +278,7 @@ module('unit/projection', function(hooks) {
               title: `Mr. Popper's Penguins`
             },
             meta: {
-              projectionTypes: [BOOK_EXCERPT_PROJECTION_CLASS_PATH]
+              partial: true,
             }
           },
         ]
@@ -377,7 +377,7 @@ module('unit/projection', function(hooks) {
             id: BOOK_ID,
             type: BOOK_CLASS_PATH,
             meta: {
-              projectionTypes: [BOOK_EXCERPT_PROJECTION_CLASS_PATH],
+              partial: true,
             },
             attributes: {
               title: BOOK_TITLE
@@ -606,9 +606,6 @@ module('unit/projection', function(hooks) {
               'chapter-1': NEW_CHAPTER_TEXT,
               description: NEW_DESCRIPTION,
             },
-            meta: {
-              projectionTypes: [BOOK_CLASS_PATH]
-            }
           },
         });
       });
@@ -666,7 +663,7 @@ module('unit/projection', function(hooks) {
                */
               // description: NEW_DESCRIPTION,
             }
-          }]
+          }],
         });
       });
 
@@ -882,8 +879,8 @@ module('unit/projection', function(hooks) {
               author: {
                 location: NEW_AUTHOR_LOCATION,
                 age: NEW_AUTHOR_AGE,
-              }
-            }
+              },
+            },
           },
         });
       });
@@ -989,11 +986,6 @@ module('unit/projection', function(hooks) {
       } = this.watchers;
       let baseCounters = baseRecordWatcher.counters;
       let excerptCounters = excerptWatcher.counters;
-
-      // run(() => {
-      //   set(projectedExcerpt, 'author.location', NEW_AUTHOR_LOCATION);
-      //   set(projectedExcerpt, 'author.age', NEW_AUTHOR_AGE);
-      // });
 
       assert.watchedPropertyCount(baseCounters['author.age'], 1, 'Afterwards we have dirtied excerpt.author.age');
       assert.watchedPropertyCount(excerptCounters['author.age'], 1, 'Afterwards we have dirtied excerpt.author.age');
@@ -1399,14 +1391,20 @@ module('unit/projection', function(hooks) {
         store.push({
           data: {
             id: BOOK_ID,
-            type: BOOK_CLASS_PATH,
-            meta: {
-              // TODO Do we want this partial or not?
-              partial: true,
-            },
+            type: BOOK_EXCERPT_PROJECTION_CLASS_PATH,
             attributes: {}
           },
           included: [
+            {
+              id: BOOK_ID,
+              type: BOOK_CLASS_PATH,
+              meta: {
+                partial: true,
+              },
+              attributes: {
+                publisher: PUBLISHER_ID
+              }
+            },
             {
               id: PUBLISHER_ID,
               type: PROJECTED_PUBLISHER_CLASS,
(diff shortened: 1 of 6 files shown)


tests/unit/projection-test.js
```diff
@@ -711,30 +718,21 @@ module('unit/projection', function(hooks) {
                 age: AUTHOR_AGE,
               },
             },
-            meta: {
-              projectionTypes: [BOOK_CLASS_PATH]
```
runspired: This still has value because when it is present it alerts us that this data is partial data, we otherwise do not have a way of determining this. I was planning on using it's presence for developing the `load-data-but-leave-in-unloaded-state` behavior.

tests/unit/projection-test.js
```diff
@@ -711,30 +718,21 @@ module('unit/projection', function(hooks) {
...
-            }
           },
         });
 
         projectedExcerpt = store.push({
           data: {
             id: BOOK_ID,
-            type: BOOK_CLASS_PATH,
```
runspired: good catch

tests/unit/projection-test.js
```diff
@@ -711,30 +718,21 @@ module('unit/projection', function(hooks) {
...
-            },
+            type: BOOK_EXCERPT_PROJECTION_CLASS_PATH,
             attributes: {},
           },
         });
 
         projectedPreview = store.push({
           data: {
             id: BOOK_ID,
-            type: BOOK_CLASS_PATH,
```
runspired: another good catch :)

runspired: another good catch :)

tests/unit/projection-test.js
```diff
@@ -955,7 +953,9 @@ module('unit/projection', function(hooks) {
       assert.equal(get(projectedExcerpt, 'author.age'), AUTHOR_AGE, 'excerpt has the correct author.age');
     });
 
-    test('Updating an embedded object property on a projection updates the base-record and other projections', function(assert) {
+    // Skipped because we cannot really simulate an update to a projection, it is always an update to the base record
+    // only fetches are something we can distinguish
+    skip('Updating an embedded object property on a projection updates the base-record and other projections', function(assert) {
```
runspired: I would unskip this, we can simulate an update to a projection if `projectionTypes` is our way of signaling `partial-data`. The payload should be update however to push the empty-projection as the primary record and the partial book in `included`

tests/unit/projection-test.js
```diff
@@ -998,7 +998,7 @@ module('unit/projection', function(hooks) {
       assert.equal(get(projectedExcerpt, 'author.age'), NEW_AUTHOR_AGE, 'excerpt has the correct author.age');
     });
 
-    test('Updating an embedded object property on a nested projection updates the base-record and other projections', function(assert) {
+    skip('Updating an embedded object property on a nested projection updates the base-record and other projections', function(assert) {
```
runspired: same as above RE unskip

tests/unit/projection-test.js
```diff
@@ -1084,9 +1084,6 @@ module('unit/projection', function(hooks) {
             type: BOOK_CLASS_PATH,
             attributes: {
               publisher: `urn:${PUBLISHER_CLASS}:${PUBLISHER_ID}`,
-            },
-            meta: {
-              projectionTypes: [BOOK_CLASS_PATH]
```
runspired: keep projectionTypes

tests/unit/projection-test.js
```diff
@@ -1097,32 +1094,27 @@ module('unit/projection', function(hooks) {
                 name: PUBLISHER_NAME,
                 location: PUBLISHER_LOCATION,
                 owner: PUBLISHER_OWNER,
-              },
-              meta: {
-                projectionTypes: [PUBLISHER_CLASS, NORM_PROJECTED_PUBLISHER_CLASS]
```
runspired: keep projectionTypes

tests/unit/projection-test.js
```diff
@@ -1097,32 +1094,27 @@ module('unit/projection', function(hooks) {
...
               }
+            }, {
+              id: PUBLISHER_ID,
+              type: NORM_PROJECTED_PUBLISHER_CLASS,
+              attributes: {}
```
runspired: The idea here is that the normalized payload would be roughly

```
{
  data: <TOP_LEVEL_PROJECTION>,
  included: [
    <PARTIAL_PRIMARY_RECORD>,
    <PARTIAL_RESOLVED_RECORD>,
    <RESOLVED_PROJECTION (empty)>
  ]
}
```

confirm?

tests/unit/projection-test.js
```diff
@@ -1097,32 +1094,27 @@ module('unit/projection', function(hooks) {
...
           ]
         });
 
         projectedExcerpt = store.push({
           data: {
             id: BOOK_ID,
-            type: BOOK_CLASS_PATH,
-            meta: {
-              projectionTypes: [BOOK_EXCERPT_PROJECTION_CLASS_PATH],
-            },
```
runspired: 👍

tests/unit/projection-test.js
```diff
@@ -1097,32 +1094,27 @@ module('unit/projection', function(hooks) {
...
+            attributes: {}
           },
         });
 
         projectedPreview = store.push({
           data: {
             id: BOOK_ID,
-            type: BOOK_CLASS_PATH,
-            meta: {
-              projectionTypes: [BOOK_PREVIEW_PROJECTION_CLASS_PATH],
```
runspired: 👍

tests/unit/projection-test.js
```diff
@@ -1262,26 +1254,13 @@ module('unit/projection', function(hooks) {
...
-              type: PUBLISHER_CLASS,
-              attributes: {
-                location: NEW_PUBLISHER_LOCATION,
-                owner: NEW_PUBLISHER_OWNER
-              },
-              meta: {
-                projectionTypes: [PUBLISHER_CLASS]
-              }
-            }
-          ]
```
runspired: semantically this is probably the same change as just dropping the erroneous inclusion of `projectionTypes`; however, since the idea is to simulate an included record in a different payload I'd prefer to keep it structured that way.

tests/unit/projection-test.js
```diff
@@ -1307,6 +1286,7 @@ module('unit/projection', function(hooks) {
 
       run(() => {
         set(projectedExcerpt, 'publisher.location', NEW_PUBLISHER_LOCATION);
+        set(projectedExcerpt, 'publisher.owner', NEW_PUBLISHER_OWNER);
```
runspired: I believe this should be an expect throws and the counts below should reflect that owner does not change since the publisher projection does not allow `owner`.

tests/unit/projection-test.js
```diff
@@ -1352,7 +1332,7 @@ module('unit/projection', function(hooks) {
       assert.equal(get(projectedExcerpt, 'publisher.owner'), PUBLISHER_OWNER, 'excerpt has the correct publisher.owner');
     });
 
-    test('Updating a resolution property via a projection updates the base-record, other projections and nested projections', function(assert) {
+    skip('Updating a resolution property via a projection updates the base-record, other projections and nested projections', function(assert) {
```
runspired: unskip as with above

tests/unit/projection-test.js
```diff
@@ -1363,26 +1343,13 @@ module('unit/projection', function(hooks) {
       run(() => {
         store.push({
           data: {
-            id: BOOK_ID,
```
runspired: I believe the correct change here is to move to the following format:

```
{
  data: <TOP_LEVEL_PROJECTION>,
  included: [
    <PARTIAL_PRIMARY_RECORD>,
    <PARTIAL_RESOLVED_RECORD>,
    <RESOLVED_PROJECTION (empty)>
  ]
}
```

tests/unit/projection-test.js
```diff
@@ -1400,7 +1367,7 @@ module('unit/projection', function(hooks) {
       assert.equal(get(projectedExcerpt, 'publisher.owner'), NEW_PUBLISHER_OWNER, 'excerpt has the correct publisher.owner');
     });
 
-    test('Updating a resolution property via a nested projection updates the base-record, other projections', function(assert) {
+    skip('Updating a resolution property via a nested projection updates the base-record, other projections', function(assert) {
```
runspired: also unskip

tests/unit/projection-test.js
dnachev: Should we just have a flag `partial: true`?

runspired: Might be easier yes

tests/unit/projection-test.js
dnachev: Actually, you are correct. I will check the changes again to ensure I haven't just placed the base record as a top level, when it should be in included.

tests/unit/projection-test.js
dnachev: So, the idea is to always have book/projection of a book in the `data` and publisher in the `included`?

I was just trying to push the minimal required data to exercise the use case, in which case is an update to the publisher.

tests/unit/projection-test.js
dnachev: That's fair, didn't look at whether we have failing `set` for resolved entity.

tests/unit/projection-test.js
```diff
@@ -619,22 +624,24 @@ module('unit/projection', function(hooks) {
     });
 
     test('Setting a projection updates the base-record and other projections', function(assert) {
-      let excerpt = this.records.projectedExcerpt;
+      let preview = this.records.projectedPreview;
```
runspired: why this change?

tests/unit/projection-test.js
runspired: I mistook excerpt here for preview, this set should not be expected to throw

tests/unit/projection-test.js
dnachev: The excerpt doesn't whitelist the `chapter-1` property, which breaks. I figured `preview` suits the rest of the test conditions.

PR #34, opened 2017-11-17
[FEAT] support for partially loaded state, findRecord/peekRecord/peekAll

runspired:
This PR requires #33 which it currently includes

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -19,6 +19,8 @@ const {
   }
 } = RootState;
 
+const FakeAttributeMeta = { isAttribute: true };
+
 class EmbeddedSnapshot {
   constructor(record) {
     this.record = record;
@@ -227,6 +229,10 @@ export default class MegamorphicModel extends Ember.Object {
         this.setUnknownProperty(key, value);
       }
     }
+  }
+
+  static metaForProperty() {
+    return FakeAttributeMeta;
   }
 
   static get isModel() {
--- a/addon/model.js
+++ b/addon/model.js
@@ -207,6 +207,7 @@ export default class MegamorphicModel extends Ember.Object {
   }
 
   initBaseModel(baseModelName) {
+    // TODO document that we are dependent on store.push pushing in included first which has the base record data in it
     let baseModel = this._store.recordForId(baseModelName, this.id);
     // TODO We can probably do it with an observer although it introduces an async behavior
     // TODO Don't forget to remove it when this M3 model has been destroyed
--- a/addon/schema-manager.js
+++ b/addon/schema-manager.js
@@ -12,9 +12,10 @@ export class SchemaManager {
   }
 
   computeBaseModelName(projectionModelName) {
-    if (typeof this.schema.computeBaseModelName !== 'function') {
-      return null;
+    if (!this.schema || typeof this.schema.computeBaseModelName !== 'function') {
+      return;
     }
+
     return this.schema.computeBaseModelName(projectionModelName);
   }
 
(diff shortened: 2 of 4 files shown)


addon/schema-manager.js
```diff
@@ -12,9 +12,10 @@ export class SchemaManager {
   }
 
   computeBaseModelName(projectionModelName) {
-    if (typeof this.schema.computeBaseModelName !== 'function') {
-      return null;
+    if (!this.schema || typeof this.schema.computeBaseModelName !== 'function') {
```
dnachev: None of the other functions checks whether `this.schema` is set, which is not a terrible thing, but makes this function stand out... kind of.

addon/model.js
```diff
@@ -259,6 +259,7 @@ export default class MegamorphicModel extends Ember.Object {
   }
 
   initBaseModel(baseModelName) {
+    // TODO document that we are dependent on store.push pushing in included first which has the base record data in it
```
dnachev: That's actually the beauty of `recordForId` - we are not dependent, although we will not have the data until it has been pushed, but this should be fine.

addon/schema-manager.js
```diff
@@ -60,6 +61,10 @@ export class SchemaManager {
     return transform ? transform(value) : value;
   }
 
+  registerAsyncSchemas(schemas) {
```
dnachev: I'm not sure whether we want SchemaManager to be involved in asynchronous schemas - this is something we are coming up with and can be isolated in our specific implementation of the schema. The Schema Manager will always ask for fresh information irrespective.

runspired: We do because nothing gets direct access to the schema, the manager is really just a singleton instance proxy

dnachev: In order to support dynamic schemas we probably need the schema to be a service, so we can throw it away between app instances. But the point I'm trying to make is this is not something, which the `SchemaManager` should be concerned with, IMHO. It is specific to the application and the particular usage of m3, how this is handled.

runspired: The schema manager has the same API as the schema for a reason.. so that we can throw away the schema instance between test and application instances but maintain a stable API reference throughout the code..

addon/schema-manager.js
hjdivad: do not silently tolerate `undefined` schema

addon/schema-manager.js
hjdivad: The schema manager is the internal API for interacting with the user's schema, but the schema object itself is userland.

The schema manager should have a minimal API (but it's fine for the userland schema to have a broader API).  For the schema manager to have an API that is not called internally is a smell.

It does not look to me that the schema manager needs to know about asynchronously loaded schemas.  What's the case where it does?

tests/unit/projection-test.js
```diff
@@ -94,308 +140,328 @@ module('unit/projection', function(hooks) {
...
-          attributes: ['title', 'author', 'chapter-1', 'year', 'publisher'],
-        },
-        [PUBLISHER_CLASS]: {},
-        // this schema must come with the parent schema
-        [NORM_PROJECTED_AUTHOR_CLASS]: {
-          attributes: ['location', 'name']
-        },
-        [NORM_PROJECTED_PUBLISHER_CLASS]: {
-          projectedType: PUBLISHER_CLASS,
-          attributes: ['location', 'name']
```
hjdivad: why is this moved to a closure var?

runspired: because I didn't realize we bound the scope for methods in the schema and so configured it this way. The change happens because @dnachev also added in accessing of models in a separate PR which when rebased became a change vs an addition. I can revert the use of a closure var now that I see that schemas have a context.

addon/schema-manager.js
runspired: @hjdivad so you're saying that in user-land registering an async-schema would be handled by an end-user directly manipulating `schema.models` for instance ?

addon/schema-manager.js
runspired: 👍

PR #35, opened 2017-11-17
Migrate to merge strategy from overwrite strategy

dnachev:
The migration is not ready yet, there are still some tests failing mainly for cases, which change the type of properties (primitive -> model, model -> primitive) etc.

But I've adjusted the existing tests for detecting deleted properties and split them in tests for validating omitted properties do not trigger changes and nulled properties trigger changes. These are worth reviewing.

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -333,7 +333,7 @@ export default class MegamorphicModel extends Ember.Object {
 
         let oldIsRecordArray = oldValue && oldValue.constructor === M3RecordArray;
         let oldWasModel = oldValue && oldValue.constructor === MegamorphicModel;
-        let newIsObject = newValue !== null && isObject(newValue);
+        let newIsObject = isObject(newValue);
 
         if (oldWasModel && newIsObject) {
           oldValue._didReceiveNestedProperties(internalModel._data[key], nestedKeys);
--- a/addon/model.js
+++ b/addon/model.js
@@ -4,10 +4,10 @@ import { dasherize } from '@ember/string';
 
 import SchemaManager from './schema-manager';
 import M3RecordArray from './record-array';
-import { setDiff, OWNER_KEY, merge } from './util';
+import { OWNER_KEY, merge } from './util';
 
 const {
-  changeProperties, get, set, isEqual, propertyWillChange, propertyDidChange, computed, A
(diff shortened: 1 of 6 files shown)


dnachev:
OK, all tests should pass now and this should be ready for merge.

dnachev:
Changed the `todo` tests to be `skip` - it seems the command line reporter doesn't handle `todo` tests.

addon/model.js
```diff
@@ -307,17 +255,13 @@ export default class MegamorphicModel extends Ember.Object {
     return this._baseModel != null;
   }
 
-  get _updateStrategy() {
-    // TODO There is no good way to determine which strategy to use, so using some heuristics for now
-    return MERGE_STRATEGY;
```
runspired: it appears that @hjdivad had the intention of making the update strategy potentially pluggable, should we keep that plumbing?

addon/model.js
```diff
@@ -450,7 +390,7 @@ export default class MegamorphicModel extends Ember.Object {
   }
 
   _changedKeys(data) {
```
runspired: I know this was pre-existing but this method name is unclear, perhaps `_updateChangedKeys` would be better.

addon/util.js
```diff
@@ -25,12 +25,16 @@ export const OWNER_KEY = (function() {
   }
 })();
 
-function isObject(value) {
+export function isObject(value) {
```
runspired: should we just add the `!null` check here?

tests/unit/model-test.js
```diff
@@ -1007,11 +1043,13 @@ test('omitted attributes in nested models are treated as deleted', function(asse
           name: `Harry Potter and the Sorcerer's Stone`,
           nextChapter: {
             id: 'ch1',
-            number: 1,
+            name: null,
```
runspired: `name` here appears in a redundant position

runspired: (and could be muddying the test?)

addon/model.js
dnachev: I think he wanted this to be pluggable in Ember Data, not necessarily in M3, but I will let him comment.

addon/model.js
dnachev: This maps to the `InternalModel._changedKeys`. Apart from this, I don't have preference for the name.

PR #36, opened 2017-11-18
Cleaning up the notifications

dnachev:
This is building on top of fixing the merge strategy PR #35.

With slight inefficiency, the notifications can be changed to use single mechanism to invalidate properties on set and on update from the store.

Using the `changedKeys` structure from PR #35 we can express the changes in the nested model, so `_notifyProperties` can walk the object graph. In case of the top level model, this change becomes simple `_notifyProperties` for the single changed key. If the model doesn't have projections, then we resort to the old simple behavior ensuring we don't penalize the non-projection use cases.


addon/model.js
```diff
@@ -159,11 +159,29 @@ function disallowAliasSet(object, key, value) {
...
+ * // foo.bar.baz.prop1
+ * // foo.bar.baz.prop2
+ * // the input should be:
+ * let path = ['foo', 'bar', 'baz'];
+ * let changedKeys = ['prop1', 'prop2'];
+ *
+ * // the resulting list should be:
+ * let result = constructChangedKeys(path, changedKeys);
+ * // result is:
+ * // ['foo', ['bar', ['baz', 'prop1', 'prop2']]]
```
runspired: is this correct? I would have expected:

```
['foo', ['bar', ['baz', ['prop1', 'prop2']]]]
```

dnachev: It is a possible alternative, but the current code slices the array to get the changed key and the list of nested keys.

runspired: what would happen if `bar` had also changed `biz` in addition to `baz.prop1` and `baz.prop2` ? `['foo', ['bar', ['biz' 'baz', 'prop1', 'prop2']]]` wouldn't work.

dnachev: The list of changed properties will be:
`['foo', ['bar', ['baz', 'prop1', 'prop1'], 'biz']]` or `['foo', ['bar', 'biz', ['baz', 'prop1', 'prop1']]]`

For the record, `constructChangedKeys` cannot generate such structure as it expects a path and list of changed keys, although the changed keys, might be something constructed with another call to `constructChangedKeys`.

runspired: I see, so today we would end up with separate lists of changed keys per-path, not a union of those lists

dnachev: I would like to say 'yes', but I'm not sure it will completely right. The specific function we are discussing here deals with changed keys in single path, therefore you can say it generate separate lists per path, although it is more correct to say it generate single list always as the path is only one.

The overall structure encodes the changes without duplications, e.g. if there are multiple changed keys in single path `foo.bar`, the prefix `foo.bar` will appear once in the overall structure. However, this is only handled in the `merge` function.

runspired:
Looks as though the failures were all testem tooling related, and most scenarios completed successfully. Merging.

PR #39, opened 2017-11-21
Fix the `isDirty` property

dnachev:
The `isDirty` property doesn't actually exist in the `DS.Model`. The closest there is `dirtyType`. We should either fix it or drop it completely.


dnachev:
Hm, weird, the `isDirty` property is on the state, but not exposed through `DS.Model`.

PR #40, opened 2017-11-22
[Projection Support] Tests for delete projections

dnachev:
Requires PR #36.

This is an experiment to see what it will take to keep the delete states of a record in sync. It is missing the handling of the rollback.

Main concern is that it is quite elaborate and requires to duplicate a lot of the logic already existing in the state machine.  The alternative is to intercept `transitionTo` and invoke `transitionTo` on all related records if the state is a `deleted` substate. However, not sure how public this will be.


dnachev:
@runspired Restructured this PR to be just tests.

PR #41, opened 2017-11-28
[Projection Support] Addressing some feedback from PR #38

dnachev:
This addresses some of the feedback from PR #38 

- Grouped the projection related properties in the `init` function.
- Converted `_internalModel` to be a getter, which always delegates to the `baseModel`.
- Cleaned up the `merge` function and removed the check for `Date` in the `isObject` function.
- Removed the `modelIsProjection` function from the `SchemaManager`

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -226,18 +226,18 @@ export default class MegamorphicModel extends Ember.Object {
     this._store = properties.store;
     this.id = this._internalModel.id;
     this._internalModel = properties._internalModel;
-    this._projections = null;
 
     this._schema = SchemaManager;
 
     let baseModelName = this._schema.computeBaseModelName(this._modelName);
     this._baseModel = baseModelName ? this.initBaseModel(baseModelName) : null;
+    this._projections = null;
+    this._path = this._path || [];
 
     this._cache = Object.create(null);
 
     this._topModel = this._topModel || this;
     this._parentModel = this._parentModel || null;
-    this._path = this._path || [];
     this._init = true;
 
     this._flushInitProperties();
@@ -290,7 +290,15 @@ export default class MegamorphicModel extends Ember.Object {
   }
 
   get _modelName() {
-    return this._internalModel.modelName;
+    return this.__internalModel.modelName;
+  }
+
+  get _internalModel() {
+    return this.baseModel ? this.baseModel._internalModel : this.__internalModel;
+  }
+
+  set _internalModel(internalModel) {
+    this.__internalModel = internalModel;
   }
 
   __defineNonEnumerable(property) {
@@ -565,7 +573,6 @@ export default class MegamorphicModel extends Ember.Object {
 }
 
 MegamorphicModel.prototype.store = null;
-MegamorphicModel.prototype._internalModel = null;
 MegamorphicModel.prototype._parentModel = null;
 MegamorphicModel.prototype._topModel = null;
 MegamorphicModel.prototype._path = null;
--- a/addon/model.js
+++ b/addon/model.js
@@ -290,15 +290,17 @@ export default class MegamorphicModel extends Ember.Object {
   }
 
   get _modelName() {
-    return this.__internalModel.modelName;
+    return this._internalModel.modelName;
   }
 
-  get _internalModel() {
-    return this.baseModel ? this.baseModel._internalModel : this.__internalModel;
+  get _data() {
+    let internalModel = this._baseModel ? this._baseModel._internalModel : this._internalModel;
+    return internalModel._data;
   }
 
-  set _internalModel(internalModel) {
-    this.__internalModel = internalModel;
+  set _data(data) {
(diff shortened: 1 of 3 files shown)


runspired:
@dnachev does this need to point at the `cleanup-notifications` branch or can it be safely pointed straight into the m3-projections branch instead?

dnachev:
@runspired Apparently, simple rebase is not enough. I haven't looked deeper.

addon/model.js
```diff
@@ -24,7 +24,7 @@ const FakeAttributeMeta = { isAttribute: true };
 class EmbeddedSnapshot {
   constructor(record) {
     this.record = record;
-    this.modelName = this.record._internalModel.modelName;
+    this.modelName = this.record.modelName;
```
runspired: shouldn't this either be `this.modelName = this.record._modelName;` or M3._modelName be renamed to M3.modelName ?

dnachev: Good catch! Thinking about it, the change was supposed to not use `_internalModel`, because it will refer to the base model, but `EmbeddedSnapshot` is only used for nested models, which doesn't have `_baseModel`, therefore the line should work as it is. I will revert it.

dnachev:
@runspired Addressed your comment.

addon/model.js
hjdivad: should add a test for this

runspired:
@rwjblue you have any ideas why we're seeing these unexpected browser crashes?

addon/util.js
```diff
@@ -26,7 +26,7 @@ export const OWNER_KEY = (function() {
 })();
 
 export function isObject(value) {
```
runspired: I'm not sure I like the name `isObject` for this util anymore, maybe `isExtantNonArrayObject` ?

addon/model.js
```diff
@@ -290,7 +290,15 @@ export default class MegamorphicModel extends Ember.Object {
   }
 
   get _modelName() {
-    return this._internalModel.modelName;
+    return this.__internalModel.modelName;
```
runspired: why wouldn't this go through the `_internalModel` codepath ?

addon/model.js
```diff
@@ -290,7 +290,15 @@ export default class MegamorphicModel extends Ember.Object {
...
+  }
+
+  get _internalModel() {
+    return this.baseModel ? this.baseModel._internalModel : this.__internalModel;
+  }
+
+  set _internalModel(internalModel) {
+    this.__internalModel = internalModel;
```
runspired: Are we sure `get _internalModel() {}` should prefer `baseModel._internalModel` when present instead of `__internalModel` ? otherwise it seems we risk trolling ourselves by setting an _internalModel that we don't access back.

runspired:
It appears the test crashes are a known problem and Travis has reverted a change on their end. If the issue persists much longer we can configure chrome via `testem.js` to use `--no-sandbox` or set `sudo: required` in `.travis.yml`.

https://github.com/travis-ci/travis-ci/issues/8836

addon/util.js
dnachev: I had to see what "Extant" means :)

If you elaborate what is your concern of this being named `isObject`, we may be able to find better name.

dnachev:
I've replaced the `_internalModel` getter to be `_data` as we actually care about the base model's `_data` and not the actual `_internalModel`. This also removed the couple of questionable references and ambiguities, which @runspired pointed out.

addon/util.js
runspired: Extant basically just means "exists" or in this case "not null". My main thing is that `isObject` usually does not also check `!isArray` so I'd like to make that clear. And I figured if we are going to go so far as `isNonArrayObject` I might as well slip in a `not array and not null` point to it.

runspired: I'd also be happy with `isNonArrayObject` though.

hjdivad: `Extant` is not quite right.  It has a strong connotation of "still around" which is not really what we're testing here.

`isObject` seems ok if we change the implementation to something like `value && value.constructor === Object`

Basically we're looking for POJOs.

dnachev: I like changing it to `value.constructor === Object`. Any objections? @runspired @hjdivad

rwjblue: Checking constructor prevents using Object.create(null) objects, will that be an issue for us?

Do we care about anything other than `typeof obj === ‘object’ && obj !== null`?

runspired: We care that it's not an array. David and I discussed this OOB this evening and think we should just rename the existing implantation (not null not array type object) to something like "isEmbeddedObject" or some such

runspired: Basically rename to the reason for the checks instead of what the check is

dnachev: OK, renamed the function to `isEmbeddedObject`, because:
- embedded is already the term we use to refer to the sub-models, which can be part of a model/record
- we use the function in two different places (not in the `merge` function) and although they are similar uses (both determining whether the logic should recurse), they are different to make it difficult to find a name, which fits both cases.

PR #47, opened 2017-11-29
Cleanup tests and SchemaManager based on code-review

runspired:
Also required altering our `store.preloadData` implementation as it surfaced an issue with property-notification for already populated models. It is now more similar to `store._load` but keeps the record in the EmptyState and skips notifying the record-array-manager if it truly is a "preload" and not an update. cc @igorT

### DIFF
--- a/addon/initializers/m3-store.js
+++ b/addon/initializers/m3-store.js
@@ -9,6 +9,7 @@ import QueryCache from '../query-cache';
 
 const { assign, isEqual } = Ember;
 const { dasherize } = Ember.String;
+const EmptyState = new InternalModel().currentState;
 
 // TODO: this is a stopgap.  We want to replace this with a public
 // DS.Model/Schema API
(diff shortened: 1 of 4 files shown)


tests/unit/projection-test.js
```diff
@@ -140,19 +83,60 @@ module('unit/projection', function(hooks) {
...
+          },
+          defaults: {
+            publisher: 'Penguin Classics',
+            hardback: true,
+            paperback: true,
+            publishedIn: 'US',
+          },
+          transforms: {
+            pubDate(value) {
+              return new Date(Date.parse(value));
```
dnachev: Do we actually use this anywhere?

addon/initializers/m3-store.js
```diff
@@ -55,17 +56,15 @@ export function extendStore(Store) {
...
-      if (typeof data.relationships === 'object' && data.relationships !== null) {
-        Object.assign(preloadData, data.relationships);
-      }
+      internalModel.setupData(data);
 
-      internalModel.preloadData(preloadData);
+      if (isUpdate === true) {
+        this.recordArrayManager.recordDidChange(internalModel);
+      } else {
+        internalModel.currentState = EmptyState;
```
dnachev: Do you need to explicitly set the state here? I assumed it is correctly preinitialized.

tests/unit/projection-test.js
runspired: not sure we do

addon/initializers/m3-store.js
runspired: when actually pre-loading (e.g. the non-update case), `setupData` will move the model into a `loaded` state, so we must reset

tests/unit/projection-test.js
runspired: removed along with some other cruft

addon/initializers/m3-store.js
dnachev: Ah, yeah, forgot about this required hack :)

runspired:
Travis failure was a tooling issue and nothing to do with the new commit.

PR #49, opened 2017-11-30
Add prettier and reformat everything

dnachev:
As we discussed, added prettier with simple config, reformatted the sources and added a pre-commit hook to ensure prettier is always applied.

Any pending *commit* requires:
- add prettier
`yarn add prettier --dev --exact`
- format everything
`node_modules/.bin/prettier --single-quote --trailing-comma es5 --ignore-path .gitignore --write 'app/**/*.js' 'addon/**/*.js' 'tests/**/*.js'`

Before attempting to rebase on master.

Check it out and let me know what do you think.

### DIFF
--- a/addon/query-cache.js
+++ b/addon/query-cache.js
@@ -5,11 +5,10 @@ const { get, RSVP: { resolve } } = Ember;
 import MegamorphicModel from './model';
 import M3RecordArray from './record-array';
 
-
 function stripSlash(str, stripLeading, stripTrailing) {
   let startSlash = stripLeading && str.charAt(0) === '/';
   let endSlash = stripTrailing && str.charAt(str.length - 1) === '/';
-  return str.slice(startSlash ? 1 : 0, endSlash ? - 1 : str.length);
+  return str.slice(startSlash ? 1 : 0, endSlash ? -1 : str.length);
 }
 
 export default class QueryCache {
@@ -21,7 +20,17 @@ export default class QueryCache {
     this.__adapter = null;
   }
 
-  queryURL(url, { params=null, method='GET', cacheKey=null, reload=false, backgroundReload=false }={}, array) {
+  queryURL(
+    url,
+    {
+      params = null,
+      method = 'GET',
+      cacheKey = null,
+      reload = false,
+      backgroundReload = false,
+    } = {},
+    array
+  ) {
     let options = {};
     if (params) {
       options.data = params;
@@ -32,20 +41,28 @@ export default class QueryCache {
     let loadPromise;
 
     if (backgroundReload || reload || cachedValue === undefined) {
-      loadPromise = this._adapter.ajax(
-        adapterUrl,
-        method,
-        options
-      ).then(rawPayload => {
-        let serializer = this._store.serializerFor('-ember-m3');
-        let payload = serializer.normalizeResponse(this._store, MegamorphicModel, rawPayload, cacheKey, 'query-url');
-        let result = this._createResult(payload, { url, params, method, cacheKey }, array);
+      loadPromise = this._adapter
+        .ajax(adapterUrl, method, options)
+        .then(rawPayload => {
+          let serializer = this._store.serializerFor('-ember-m3');
+          let payload = serializer.normalizeResponse(
+            this._store,
+            MegamorphicModel,
+            rawPayload,
+            cacheKey,
+            'query-url'
+          );
+          let result = this._createResult(
+            payload,
+            { url, params, method, cacheKey },
+            array
+          );
 
-        if (cacheKey) {
-          this._addResultToCache(result, cacheKey);
-        }
-        return result;
-      });
+          if (cacheKey) {
+            this._addResultToCache(result, cacheKey);
+          }
+          return result;
+        });
     }
 
     if (reload || cachedValue === undefined) {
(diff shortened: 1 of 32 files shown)


rwjblue:
> pre-commit hook

I kinda hate these 😡 .  I think we would be better served to hook this into `.eslintrc.js` and use linting to enforce prettier in CI.

rwjblue:
For example: https://github.com/emberjs/ember-test-helpers/blob/master/.eslintrc.js

dnachev:
Removed the pre-commit hook and configured ESLint.

addon/query-cache.js
```diff
@@ -32,20 +41,28 @@ export default class QueryCache {
...
-        let result = this._createResult(payload, { url, params, method, cacheKey }, array);
-
-        if (cacheKey) {
-          this._addResultToCache(result, cacheKey);
-        }
-        return result;
-      });
+      loadPromise = this._adapter
+        .ajax(adapterUrl, method, options)
+        .then(rawPayload => {
```
hjdivad: need to configure prettier to use leading dot

dnachev:
Hm, fixed.

addon/schema-manager.js
```diff
@@ -26,14 +26,18 @@ export class SchemaManager {
 
   getDefaultValue(modelName, keyName) {
     let defaults = this._modelSchemaProperty(modelName, 'defaults');
-    if (!defaults) { return; }
+    if (!defaults) {
+      return;
+    }
```
hjdivad: this is unfortunate

tests/pages/index.js
```diff
@@ -14,21 +14,22 @@ class PageObject {
...
-    return this.$('.comment-body').
-      text().
-      replace(/\s+/g, ' ').
-      replace(/^\s*/, '').
-      replace(/\s*$/, '');
+    return this.$('.comment-body')
+      .text()
+      .replace(/\s+/g, ' ')
+      .replace(/^\s*/, '')
+      .replace(/\s*$/, '');
```
hjdivad: this sucks

tests/pages/index.js
```diff
@@ -14,21 +14,22 @@ class PageObject {
...
-      map(x =>
-        x.replace(/\s+/g, ' ').
-        replace(/^\s*/, '').
-        replace(/\s*$/, '')
+    return this.$('.comment-parts li')
+      .map((x, y) => self.jQuery(y).text())
+      .toArray()
+      .map(x =>
+        x
+          .replace(/\s+/g, ' ')
```
hjdivad: what has happened here?

seems strange to break like this when it's not necessary

hjdivad:
ok so

1. this is great, definitely like having automated formatting
2. some of the default config isn't great, if possible we should see if it could be fixed, to wit
- leading `.` sucks (although it seems there is no config for trailing `.` at the moment)
- expanding braces for trivial one-line early returns is an unfortunate waste of space.  If we can somehow configure that away it would be nice, but if not, oh well
- that arrow expressions with method chains results in an immediate break leads to some really weird expressions for single-var arguments: see https://github.com/hjdivad/ember-m3/blob/ded129bc3eeeef3c6f12e5b0c76b1a85fb498706/tests/pages/index.js#L28-L33

also it looks like the git precommit hook is gone, which is good.  We are definitely not adding a precommit hook ^_^

hjdivad:
none of the issues are significant enough to block the pr though

dnachev:
@hjdivad you have to choice the config from the following:
what is the formatting width
whether to add semicolons
use single or double quotes
indention space
whether to add trailing commas
use tabs instead of spaces

I think the prettier philosophy is to not allow configs and achieve something like the Go's community where there is single format for all Go code.

tests/pages/index.js
runspired: @hjdivad the newline or the leading dot? Personally highly prefer the leading dot to the trailing dot (also pretty sure leading dot is overwhelmingly more common a style). Always putting things on new-lines though... 🤔

tests/pages/index.js
runspired: yeah this seems bad

tests/pages/index.js
hjdivad: The leading dot.

From my experience leading dot is slightly more popular than trailing dot.

Trailing dot is just better though: they both look about the same, leading dot has no benefits, trailing dot has two benefits, one minor and one significant.

The minor benefit is that it's more general because it doesn't rely on the language having as much lookahead.  This means that trailing dot works in more languages and so you don't have to switch styles if you switch between languages.

There's a related and more significant problem, which is that the newlines will be interpreted as statement delimiters when pasting code into a REPL.  It's pretty annoying that leading-dot code cannot be copy&pasted into a REPL.

---

What's your reason for preferring leading dot?  Is there a reason beyond aesthetics?

---

In any case it seems we can't configure prettier to use trailing dot.  As I said in my earlier comment, this isn't a blocker.  I'd rather have worse but automatic formatting, than manual formatting.

hjdivad:
@dnachev yep that's a pity but as I said above, I'd still rather have automatically formatted code than manually formatted code.

---

I've restarted the builds since the failures looked to be just from the SETUID chrome sandbox issue.

PR #50, opened 2017-12-01
[Projection Support] Unload tests

dnachev:
Here are is an expansion of the two unload tests we had. Hopefully I stayed `true` to the spirit of the test structure.

Only real question is how much we want to test that unloading a related record didn't destroy internal structures, which a live record might need.

Depending on our implementation we may need a test to confirm that the underlying structure has been cleaned up of references once the final projection has been unloaded, but without an actual implementation, I don't think it is possible to write a test for it.


runspired:
Rekicked Travis to get around the headless chrome errors.

PR #51, opened 2017-12-05
[Projection Support] Write Tests

dnachev:
A starting point for the write cases discussion. I think writing projection is not as straight-forward, I expect to have some healthy discussion here.

The current thought process was the following:
- projections without IDs are not linked in any way, e.g. you can create as many as you want and modify as much as you want
- created projections with IDs are immediately linked, e.g. they represent the same underlying entity, even though this entity doesn't exist on the server
- saving a new projection record can create an entity on the server and the projection will be correctly setup to be updated by future fetches
- changing a projection dirties all associated entities (if they contain the attribute; note - the test needs to be enhanced to cover non-whitelisted properties)
- saving a projection un-dirties only the saved projection and any projection which have all of its properties saved. projections, which have properties, which have not been sent to the server continue to be dirty

I will continue enhancing these, but I also wanted to get some initial feedback about how we should be handling the write cases. One major obstacle in implementing the above rules is the fact M3 doesn't really support dirtying.

Also, we can reduce the scope for now and support something much simpler for the write cases, but not quite sure what would be a good compromise.

### DIFF
--- a/tests/unit/projection-test.js
+++ b/tests/unit/projection-test.js
@@ -1808,12 +1808,6 @@ module('unit/projection', function(hooks) {
       assert.equal(get(baseModel, '_internalModel.currentState.isDirty'), true, 'The base model should still be dirty');
     });
   });
-<<<<<<< 093cbbb25bdfb2a57ae06a44839ab907660e14b0
-=======
-
-  // TL;DR we can only proxy something that has an ID
-  skip(`Saving a newly created projection doesn't mess up the state of the base record`, function() {});
->>>>>>> create and update tests
 
   skip(`eachAttribute returns only white-listed properties`, function() {});
   skip(`Creating a projection with an unloaded schema`, function() {});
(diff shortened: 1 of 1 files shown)


tests/unit/model-test.js
```diff
@@ -681,6 +681,26 @@ module('unit/model', function(hooks) {
     assert.equal(get(model, 'publisher'), 'Harper Collins, of course', 'specified value transformed');
   });
 
+  test('early set of an ID to a newly created records is allowed', function(assert) {
```
runspired: potentially I should just make this a different PR and go ahead and land it and the implementation into master. cc @hjdivad

runspired: did so here https://github.com/hjdivad/ember-m3/pull/53

tests/unit/projection-test.js
```diff
@@ -1796,6 +1808,12 @@ module('unit/projection', function(hooks) {
       assert.equal(get(baseModel, '_internalModel.currentState.isDirty'), true, 'The base model should still be dirty');
     });
   });
+<<<<<<< 093cbbb25bdfb2a57ae06a44839ab907660e14b0
+=======
```
hjdivad: looks pretty merge conflicty

runspired: you're looking at an out-of-date diff

hjdivad: such github notifications

PR #52, opened 2017-12-05
[Projection Support] Add unload test for record arrays

dnachev:
@runspired As discussed over Slack, this is a test to validate we correctly handle record arrays retrieved from attribute array reference property.

### DIFF
--- a/tests/unit/projection-test.js
+++ b/tests/unit/projection-test.js
@@ -112,10 +112,18 @@ module('unit/projection', function(hooks) {
           resolvedTypes: {
             publisher: NORM_PROJECTED_PUBLISHER_CLASS,
             author: NORM_PROJECTED_AUTHOR_CLASS,
+            otherBooksInSeries: NORM_BOOK_PREVIEW_PROJECTION_CLASS_PATH,
           },
           // if you want to project an embedded model then it must have a type
           //  computedEmbeddedType
-          attributes: ['title', 'author', 'chapter-1', 'year', 'publisher'],
+          attributes: [
+            'title',
+            'author',
+            'chapter-1',
+            'year',
+            'publisher',
+            'otherBooksInSeries',
+          ],
         },
         [PUBLISHER_CLASS]: {},
         // this schema must come with the parent schema
@@ -2023,6 +2031,8 @@ module('unit/projection', function(hooks) {
 
   module('unloading/deleting records', function(hooks) {
     const BOOK_ID = 'isbn:123';
+    const OTHER_BOOK_ID = 'isbn:456';
+    const OTHER_BOOK_URN = `urn:${NORM_BOOK_CLASS_PATH}:${OTHER_BOOK_ID}`;
     const BOOK_TITLE = 'Alice in Wonderland';
 
     hooks.beforeEach(function() {
@@ -2044,6 +2054,7 @@ module('unit/projection', function(hooks) {
             type: BOOK_CLASS_PATH,
             attributes: {
               title: BOOK_TITLE,
+              otherBooksInSeries: [OTHER_BOOK_URN],
             },
           },
         });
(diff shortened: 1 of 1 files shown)


runspired:
@dnachev I remember discussing this but I find this test and the descriptions within it very unclear. Also, I suspect there are actually two or three different scenarios here that need tested IIRC. 

1 - unloading a resolved projection contained within a record-array is possible
2 - unloading a top-level projection unloads any embedded projections
3 - something about unloading an embedded projection on its own is/is not possible (since its a subset of embedded data on the parent object)

dnachev:
overarching - when M3 have an array references (`computeAttributeReference`), it utilizes a record array to return the list of resolved records. The test is validating these record arrays behave correctly.

1 - the record array should be referencing normal records, therefore unloading should always be possible. The test, however, validates the record is removed from the record array when it is unloaded.
The test actually validates the length of the record array changes, but this is not consistent with how it behaves if a record is missing when initially creating the record array. Both behavior can be valid depending on the use case (cc @hjdivad), but I will go ahead and align the test with the current behavior.

2 - the test doesn't validate this, but we should probably have one for this as well. Here, we really expects the embedded object to have been destroyed and nothing else, see 3) why.

3 - we cannot unload embedded records, because they are not really tracked by ED. I think David already fixed this in M3 to disallow unloading of embedded records.

runspired:
@dnachev can you rebase this onto the model-data branch?

dnachev:
Rebased, but because the tests fail for the branch ...

tests/unit/projection-test.js
```diff
@@ -2224,6 +2235,82 @@ module('unit/projection', function(hooks) {
...
+        });
+
+        assert.equal(
+          get(booksInSeriesBase, 'length'),
+          1,
+          'Expected otherBooksInSeries length to be unchanged for base'
+        );
+        assert.equal(
+          get(booksInSeriesProjectedPreview, 'length'),
+          1,
```
runspired: why don't we expect it to be removed?

tests/unit/projection-test.js
```diff
@@ -2224,6 +2235,82 @@ module('unit/projection', function(hooks) {
...
+          'Expected otherBooksInSeries length to be unchanged for projected preview'
+        );
+        assert.equal(
+          booksInSeriesProjectedPreview.getObjectAt(0),
+          null,
```
runspired: why do we expect `null` instead of `undefined`? ember-data bug?

tests/unit/projection-test.js
dnachev: Because if the record is not available (`peekRecord` returns `undefined`/`null`) in the first place, when the record array is created, it will leave a whole in the array instead of shortening the record array. This behavior is defined in the `resolveRecordArrayInternalModels` function.
We should expect the same behavior if the code forcibly unload a record and cause `peekRecord` to return `undefined`/`null`.

tests/unit/projection-test.js
dnachev: The `assert.equal` is not strict, therefore `null` or `undefined` are both valid results for the purpose of the test. Ideally, the value should be synchronized with the result from `peekRecord` in case of missing record.

tests/unit/projection-test.js
runspired: so is this the case where we know a record should exist but it simply has not been loaded?

dnachev: Well, we assume it should exist, but it does not. There are several cases, which may lead to this and the application will have to expect these case and handle them accordingly.

hjdivad: @runspired m3 doesn't support async relationships (which is more or less what this is).

The more consistent thing to do here is to assert.

tests/unit/projection-test.js
hjdivad: No, we should assert here.  The analog is ember-data sync relationships, in which case attempts to retrieve non-loaded records results in an assertion failure.

dnachev: @hjdivad The current behavior is required for the following use case:
- We fetch an array with specific projection
- We refetch the array with a different projection
- The refetch may have added new elements, which however were fetched with different projection, therefore the original references will now attempt to reference a record, which doesn't have the required projection.

The current agreement for this case is to handle it gracefully (either by the app or ember-m3) by filtering out records from the array, which doesn't have the required projections.

PR #53, opened 2017-12-05
[FEAT] disallow late ID setting

runspired:
As discussed OOB with @hjdivad, this PR asserts that you cannot `model.set('id', '1')`.

For client-side creation with ID, ID must be passed in via `store.createRecord('foo', { id: '1' });`

cc @dnachev

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -214,7 +214,6 @@ export default class MegamorphicModel extends Ember.Object {
     super.init(...arguments);
     this._store = properties.store;
     this._internalModel = properties._internalModel;
-    this.id = this._internalModel.id;
     this._cache = Object.create(null);
     this._schema = SchemaManager;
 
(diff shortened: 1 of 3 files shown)


runspired:
Travis failures are mysterious, most ember versions pass but the failed tests are essentially failing because they passed.  Will investigate.

addon/model.js
```diff
@@ -472,7 +484,6 @@ MegamorphicModel.prototype.store = null;
 MegamorphicModel.prototype._internalModel = null;
 MegamorphicModel.prototype._parentModel = null;
 MegamorphicModel.prototype._topModel = null;
-MegamorphicModel.prototype.id = null;
```
hjdivad: Why drop this from the prototype? Isn't that just going to get us down the `setUnknownProperty` path unnecessarily?

dnachev:
LGTM.

addon/model.js
runspired: It won't because `id` is now a class property.

PR #57, opened 2017-12-14
[Projection Support] Refactor nested model data

dnachev:
The refactoring changes so the top-level model data can now track its child model datas and can delegate the handling of updates to them in the proper cases. This allows the child mode datas to correctly notify their associated records of the changes in the data without resorting to complex data structure, which needs to be passed to `_notifyProperties` and has the potential to work much better with the Store Wrapper API from the model data.

The code is rather straight-forward except handling of arrays as in this case, we can have multiple model datas for single key in the top-level model. The proposed solution for this is to not track model datas in this case, because the merge updates algorithm will override the whole array and it will not attempt to merge it at all. The property will still be invalidated and re-read from the top-level model data recreating these corresponding nested model datas, therefore there is no real need to retain the nested model datas around.

The other big change this PR does is to remove the dirty tracking from the model data for M3. In order to support dirty tracking, there are some corner cases, primarily around around nested arrays (again), which will either:
- require an identifier for the array's elements to be provided
- automatically drop pending changes from models in the array, which makes for really unpredictable behavior

We can punt dealing with dirty tracking until later, when we have a better understanding of the use cases.


addon/model-data.js
```diff
@@ -1,183 +1,212 @@
...
+      return changedKeys;
+    }
+
+    let updatedKeys = Object.keys(updates);
+
+    for (let i = 0; i < updatedKeys.length; i++) {
+      let key = updatedKeys[i];
+      let newValue = updates[key];
+
+      if (isEqual(data[key], newValue)) {
```
runspired: I'm unsure that we need `isEqual` here vs `===`: https://github.com/emberjs/ember.js/blob/db34024bae2caa06c4ae845d1dc3261a02e35b69/packages/ember-runtime/lib/is-equal.js#L43-L53

PR #61, opened 2018-01-09
[Projection Support] Initial work for cache consistency and property invalidation

dnachev:
The changes are somewhat disconnected, here is an overview:
- `preloadData` monkey patch to enable us to put a base record data without marking as loaded
- Linking a projection model data to a base model data (no support for save yet)
- Some small changes here and there like relying on `setAttr` instead of assigning directly
- Unskipping relevant tests, which are now passing against Igor's Ember Data branch

### DIFF
--- a/addon/initializers/m3-store.js
+++ b/addon/initializers/m3-store.js
@@ -51,18 +51,16 @@ export function extendStore(Store) {
     _preloadSingleResource(data) {
       let modelName = dasherize(data.type);
       let modelData = this.modelDataFor(modelName, data.id);
-      let isUpdate = false;
-      // let isUpdate = modelData.currentState.isEmpty === false;
 
-      // TODO All other uses of setupData notifies the record of any
-      // properties changes, but not this one. Temporary we have an
-      // explicit parameter to make setupData notify the record as well
-      modelData.setupData(data, true, true);
+      let changedKeys = modelData.setupData(data, true, true);
 
-      if (isUpdate === true) {
-        // this.recordArrayManager.recordDidChange(modelData);
-      } else {
-        // internalModel.currentState = EmptyState;
+      // We need to explicitly notify any possible record, which may have been associated
+      // with the preloaded ModelData. We don't delegate this to InternalModel
+      // to not confuse the implementation whether the `setupData` call also means
+      // it has been loaded
+      let internalModel = this._internalModelForId(modelName, data.id);
+      if (internalModel && internalModel.hasRecord) {
+        internalModel._record._notifyProperties(changedKeys);
       }
     },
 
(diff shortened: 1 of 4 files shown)


addon/initializers/m3-store.js
```diff
@@ -18,6 +22,50 @@ export function extendStore(Store) {
...
+    _preloadSingleResource(data) {
+      let modelName = dasherize(data.type);
+      let modelData = this.modelDataFor(modelName, data.id);
+      let isUpdate = false;
+      // let isUpdate = modelData.currentState.isEmpty === false;
+
+      // TODO All other uses of setupData notifies the record of any
+      // properties changes, but not this one. Temporary we have an
+      // explicit parameter to make setupData notify the record as well
+      modelData.setupData(data, true, true);
```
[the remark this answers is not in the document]
dnachev: @runspired  I have addressed in a new commit, which also cleans up some things.

addon/initializers/m3-store.js
```diff
@@ -18,6 +22,48 @@ export function extendStore(Store) {
...
+      } else if (typeof data === 'object' && data !== null) {
+        this._preloadSingleResource(data);
+      }
+    },
+
+    _preloadSingleResource(data) {
+      let modelName = dasherize(data.type);
+      let modelData = this.modelDataFor(modelName, data.id);
+
+      let changedKeys = modelData.setupData(data, true, true);
```
runspired: we no longer need the third argument here I believe

addon/model-data.js
```diff
@@ -143,20 +159,31 @@ export default class M3ModelData {
...
    */
-  _mergeUpdates(updates, nestedCallback, calculateChanges = true) {
+  _mergeUpdates(updates, nestedCallback) {
     let data = this._data;
 
-    let changedKeys;
-    if (calculateChanges) {
-      changedKeys = [];
-    }
+    let changedKeys = [];
```
runspired: what's the effect of no longer having calculateChanges? Do we overly notify or have we guarded against that by other means?

tests/unit/projection-test.js
```diff
@@ -157,120 +155,119 @@ module('unit/projection', function(hooks) {
...
-            },
-          });
-          store.preloadData({
-            data: {
-              type: BOOK_CLASS_PATH,
-              id: FETCHED_PROJECTION_ID,
-              attributes: {
-                title: `Mr. Popper's Penguins`,
-              },
+      run(() => {
```
runspired: it seems this is mostly whitespace changes, confirm?

addon/model-data.js
dnachev: The flag is more of an optimization - there is no need to calculate changes if we don't have a record to notify. In our case, however, the algorithm already walks the changes as it needs to execute special logic for applying them, therefore computing the changes have little overhead.

tests/unit/projection-test.js
dnachev: Yes, not sure why switching from `skip` to `test` suddenly changes everything.

PR #62, opened 2018-01-09
[Projection Support] Support for saving a projection [1]

dnachev:
Although earlier PR #61 already enabled some tests related to saving,  they just happen to pass, because the default Ember Data behavior for saving works OK. However, it is not correct.

The challenge is to detect when a projection have been _created_ on the server and correctly convert the projection's ModelData to refer to the base model. Because we don't have the ID of the entity in the beginning, we cannot do this setup when the projection is first created.

The detection is done through the `setId()` function, which is called when the server sends the ID of the record. When this function is called, we need to throw away the data and link the projection ModelData to the base ModelData, which should exist by this point (because we have an ID now). 

However, we don't know how much data the server has returned back - it may have returned all the data or it may have returned just ID. To ensure we don't lose data, we need to copy any data, owned by the projection, to the base ModelData. We assume the server knows better what is the current state, therefore any pre-existing base ModelData is considered the source of truth and not overridden.

The PR also handles unloading of the base record correctly, e.g. the base record is not unloaded. However, there is no an existing hook, which can help us with this, so the code assumes additional changes to Ember Data.

Here is the list of changes required in Ember Data for this change:
- `InternalModel.setId` calls `ModelData.setId`
- `InternalModel.unloadRecord` should consult `ModelData.shouldDestroy()` before proceeding to schedule a destruction of the InternalModel.
- `InternalModel.destroy()` calls `ModelData.destroy()` as well. There is an existing `resetRecord()`, which is called when the associated record is dematerialized, but it is also called at other times and there is no guarantee that after calling `resetRecord`, the modelData will not be used again. 

@igorT @hjdivad The above functions will extend the API surface of ModelData and I will appreciate a closer look. These are the easiest things, which worked for the cases, but I'm open to more refined API changes.

### DIFF
--- a/addon/model-data.js
+++ b/addon/model-data.js
@@ -67,7 +67,7 @@ export default class M3ModelData {
 
     this.baseModelName = this._schema.computeBaseModelName(this.modelName);
 
-    if (this.baseModelName) {
+    if (this.baseModelName && this.id) {
       // TODO we may not have ID yet?
       this._initBaseModelData(this.baseModelName, id);
     } else {
@@ -121,9 +121,23 @@ export default class M3ModelData {
 
   didCommit(data) {
     let changedKeys;
-    if (data) {
+    if (data && data.attributes) {
       changedKeys = this._mergeUpdates(data.attributes, commitDataAndNotify);
       this._notifyProjectionProperties(changedKeys);
+    }
+
+    if (!this.id && data && data.id) {
+      this.id = coerceId(data.id);
+      if (this.baseModelName) {
+        // Fresh projection was saved, we need to connect it with the base model data
+        let projectionData = this._data;
+        this._initBaseModelData(this.baseModelName, this.id);
+        // TODO We only do this because there might be inflight attributes, which the server
+        // didn't include in the response
+        this.baseModelData._inverseMergeUpdates(projectionData);
+        // we need to reset the __data to reread it from the base model data
+        this.__data = null;
+      }
     }
 
     return changedKeys || [];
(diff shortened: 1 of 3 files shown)


runspired:
I think the `setId` change and `destroy` changes are necessary. I'm much less convinced about `unloadRecord` consulting `shouldDestroy`.

dnachev:
I'm not sure what is the alternative to `shouldDestroy` - if someone calls `unloadRecord` on the base record (which also holds the shared `data` hash and list of associated projections), Ember Data will remove any trace of this and the projection model datas and models will no longer have access to the data and function correctly.

We may be able to workaround this by moving the tracking to a separate objects, which we maintain on our own, then we don't need control over the InternalModel lifecycle.

dnachev:
The PR has been rebased on the latest Ember Data RFC and adapted to use the new APIs.

addon/model-data.js
```diff
@@ -121,11 +121,20 @@ export default class M3ModelData {
...
   didCommit(data) {
     let changedKeys;
-    if (data) {
+    if (data && data.attributes) {
       changedKeys = this._mergeUpdates(data.attributes, commitDataAndNotify);
       this._notifyProjectionProperties(changedKeys);
     }
 
+    let hasReceivedId = !this.id && data && data.id;
+    if (hasReceivedId) {
```
sangm: Seems like this can be `if (hasReceivedId && this.baseModelName)`

dnachev: Not quite - we always want to get the new ID (`this.id = coerceId(data.id)`) and if there is base model name, do the rest.

PR #65, opened 2018-01-12
[Projection Support] Improve handling of nested models [2]

dnachev:
The code has been refactored to correctly delegate handling updates to the nested models, but projections of the nested models are not notified of these updates, because there is no link between nested models and the normal mechanism for notifying them of changes to the data does not work.

This PR implements a way to make this work correctly for nested models projections - when a ModelData instance is asked to create a nested ModelData, it requests its own base ModelData to provide a nested ModelData, which serves as a base for the new one. At the end, there will be a structure like this:

```
+--------------+                         +----------------+
|              +                         |                |
|  Projection     ---------------------> |      Base      |
|              |                         |                |
+----------+---+                         +-------------+--+
           |                                           |
           |                                           |
        +--v-------------+                          +--v-------------+
        |                |                          |                |
        |    NestedPrj   +------------------------> |   NestedBase   |
        |                |                          |                |
        +----------+-----+                          +------------+---+
                   |                                             |
                 +-v---------------+                          +--v--------------+
                 |                 |                          |                 |
                 | NestedNestedPrj +------------------------> |NestedNestedBase
                 |                 |                          |                 |
                 +-----------------+                          +-----------------+
```
This entirely expects that all projections of a nested hash will also be presented as nested model, which is reasonable.

I don't think we have alternatives at this point, which would not conflict other decisions we've made. The one I could think of, is to run a common update strategy over the whole `data` hash and then ask each impacted `ModelData` to do its notifications. However, this conflicts with our decision where the existence of a nested model determines whether we want to merge or overwrite updates.

@igorT This is the case, which you wanted to think about. Let me know if you have something better in mind.

### DIFF
--- a/addon/model-data.js
+++ b/addon/model-data.js
@@ -75,7 +75,6 @@ export default class M3ModelData {
     } else {
       this.baseModelName = this._schema.computeBaseModelName(this.modelName);
       if (this.baseModelName && this.id) {
-        // TODO we may not have ID yet?
         this._initBaseModelData(this.baseModelName, id);
       } else {
         this.baseModelData = null;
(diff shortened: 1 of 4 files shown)


dnachev:
Superseded by #109 .

PR #66, opened 2018-01-12
[Projection Support] Do not merge updates in case of different entity

dnachev:
This is a rather straight-forward improvement over the current update strategy - if we detect a nested model entry is entirely different entity (its type or ID changed), then we should not try to merge the updates.


dnachev:
This got merged into the #71 PR.

PR #67, opened 2018-01-18
Prettier line length to 100

hjdivad:


stefanpenner:
ah, so I see your trying to impose your mega-wide monitor on all of us?

hjdivad:
@stefanpenner i only set it to 100 not 400

PR #72, opened 2018-02-05
passing _data object to schema.js

dnalagatla:
passing _data object to schema.js, in order to check the key and value and compute attribute reference.

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -54,12 +54,9 @@ function resolveValue(key, value, modelName, store, schema, model) {
     return resolvePlainArray(key, value, modelName, store, schema, model);
   }
 
-  let reference = schema.computeAttributeReference(
-    key,
-    value,
-    modelName,
-    model._internalModel._data
-  );
+  const dataHash = model._internalModel._data;
+  let reference = schema.computeAttributeReference(key, value, modelName, dataHash);
+
   if (reference) {
     if (reference.type === null) {
       // for schemas with a global id-space but multiple types, schemas may
@@ -72,7 +69,7 @@ function resolveValue(key, value, modelName, store, schema, model) {
     }
   }
 
-  let nested = schema.computeNestedModel(key, value, modelName);
+  let nested = schema.computeNestedModel(key, value, modelName, dataHash);
   if (nested) {
     let internalModel = new EmbeddedInternalModel({
       id: nested.id,
@@ -411,11 +408,14 @@ export default class MegamorphicModel extends Ember.Object {
       }
 
       let defaultValue = this._schema.getDefaultValue(this._modelName, key);
-      return (this._cache[key] = defaultValue);
+
+      // If default value is not defined, resolve the key for reference
+      if (defaultValue !== undefined) {
+        return (this._cache[key] = defaultValue);
+      }
     }
 
     let value = this._schema.transformValue(this._modelName, key, rawValue);
-
     return (this._cache[key] = resolveValue(
       key,
       value,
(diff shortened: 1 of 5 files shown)


dnachev:
@dnalagatla What about a quick test to validate it?

dnalagatla:
@hjdivad, I added unit test for verification related to this update.

tests/unit/model-test.js
```diff
@@ -713,6 +713,52 @@ module('unit/model', function(hooks) {
     assert.equal(get(model, 'hb'), true, 'alias to missing with default');
   });
 
+  test('.unknownProperty resolves id-matched values to external m3-models when special ref present in _data passed to schema', function(
```
dnachev: I would shorten the test name to `.unknownProperty passes the model data hash to schema when resolving values`.

tests/unit/model-test.js
```diff
@@ -713,6 +713,52 @@ module('unit/model', function(hooks) {
...
+      if (typeof value === 'string' && refValue && refValue === value && /^isbn:/.test(value)) {
+        return {
+          type: null,
+          id: value,
+        };
+      }
+      return null;
+    };
+
+    model.set('relatedBook', 'isbn:9780439136365');
```
dnachev: This seems to be testing that `computeAttributeReference` has a priority over a preexisting value, e.g. `computeAttributeReference` will always be consulted before using the value as is.

I generally prefer to keep the tests (especially unit tests) focused on what they are testing and nothing more.

tests/unit/model-test.js
```diff
@@ -713,6 +713,52 @@ module('unit/model', function(hooks) {
...
+            type: 'com.example.bookstore.Book',
+            attributes: {
+              name: `Harry Potter and the Order of the Phoenix`,
+            },
+          },
+        ],
+      });
+    });
+
+    SchemaManager.schema.computeAttributeReference = function(key, value, modelName, data) {
```
dnachev: They are two schools of thought for how to do this, but you should be able to change the implementation of `computeAttributeReference` at the top of the test and just have the correct fixture and asserts here.

tests/unit/model-test.js
hjdivad: there's no reason to do this `set`; it actually interferes with the test

tests/unit/model-test.js
```diff
@@ -713,6 +713,52 @@ module('unit/model', function(hooks) {
...
+      let refValue = data['*' + key];
+      if (typeof value === 'string' && refValue && refValue === value && /^isbn:/.test(value)) {
```
hjdivad: `value` will be `undefined` in this case; we should be ignoring it.

The only reason it's not `undefined` in your test is because of the `set` made on line 753.

I think you meant to `typeof refValue`

addon/model.js
```diff
@@ -54,7 +54,12 @@ function resolveValue(key, value, modelName, store, schema, model) {
     return resolvePlainArray(key, value, modelName, store, schema, model);
   }
 
-  let reference = schema.computeAttributeReference(key, value, modelName);
+  let reference = schema.computeAttributeReference(
```
hjdivad: looks great; for consistency, can we pass `_data` to the other APIs as well?

hjdivad: also please update

 - the blueprint
 - examples in readme (even if just the function args)

tests/unit/model-test.js
dnalagatla: Updated the test name.

tests/unit/model-test.js
dnalagatla: Moved the implementation of `computeAttributeReference` out of the test.

tests/unit/model-test.js
dnalagatla: Removed `set` from the test

addon/model.js
dnalagatla: I added data hash as part other APIs `isAttributeArrayReference` and `computeNestedModel`

tests/unit/model-test.js
dnalagatla: the conditional here.

tests/unit/model-test.js
dnalagatla: Removed `set` .

hjdivad:
Test failure is unrelated to this pr; something has changed in ember-beta; will need to dig into it

PR #80, opened 2018-02-10
[BUGFIX] Prevent multiple reopen calls in tests

runspired:


addon/initializers/m3-store.js
```diff
@@ -100,6 +102,11 @@ export function extendDataAdapter(DataAdapter) {
 }
 
 export function initialize() {
+  if (hasReopened === true) {
+    return;
+  }
+
+  hasReopened = true;
   extendStore(DS.Store);
```
stefanpenner: the body of this function an just be moved into the root of the module body. The following ensure the reopen happens as soon as possible, and not later when the store or adapter may already have been instantiated.


```js
import DS from 'ember-data';
extendStore(DS.Store);
extenddataAdapter(Ember.DataAdapter);

export function initialize() {}
```

PR #81, opened 2018-02-10
Clean up module-scoped init

hjdivad:
see https://github.com/hjdivad/ember-m3/pull/80#event-1467797580


hjdivad:
@rwjblue I'm 👍 with it as long as they're wrapped in functions like this so we can still unit test

rwjblue:
Aye, also FWIW the newer testing harness stuff no longer has the issue where initializers _only_ run for acceptance tests. They will now run for all types (in the newer style).

PR #86, opened 2018-02-20
Hjdivad/dirty tracking

hjdivad:
supersedes #71

### DIFF
--- a/addon/model-data.js
+++ b/addon/model-data.js
@@ -1,6 +1,9 @@
 import { isEqual } from '@ember/utils';
+import { assign, merge } from '@ember/polyfills';
+import { copy } from '@ember/object/internals';
 import { coerceId } from 'ember-data/-private';
-import { setDiff } from './util';
+
+const emberAssign = assign || merge;
 
 class M3SchemaInterface {
   constructor(modelData) {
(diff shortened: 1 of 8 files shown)


addon/model-data.js
```diff
@@ -86,46 +243,89 @@ export default class M3ModelData {
 
   setHasMany() {}
 
-  commitWasRejected() {}
+  commitWasRejected() {
```
hjdivad: probably need to recurse here as well

addon/model-data.js
```diff
@@ -142,39 +342,85 @@ export default class M3ModelData {
...
+    let changedAttributeNames = Object.keys(changedAttributes);
+    let attrs = this._attributes;
+
+    for (let i = 0, length = changedAttributeNames.length; i < length; i++) {
+      let attribute = changedAttributeNames[i];
+      let data = changedAttributes[attribute];
+      let oldData = data[0];
+      let newData = data[1];
+
+      if (oldData === newData) {
```
hjdivad: check recursive case

addon/model-data.js
```diff
@@ -142,39 +342,85 @@ export default class M3ModelData {
...
+
+        // A value in _attributes means the user has a local change to
+        // this attributes. We never override this value when merging
+        // updates from the backend so we should not sent a change
+        // notification if the server value differs from the original.
+        if (hasAttrs === true && attrs[key] !== undefined) {
+          continue;
+        }
+
+        if (!isEqual(original[key], value)) {
```
hjdivad: add comment re we will always treat array & object as different

PR #89, opened 2018-02-22
clean up some beta issues

hjdivad:
This deals with deprecations, but it doesn't handle the failures which stem from the state CPs not being treated as descriptors during dependent notifications stemming from `currentState` change events (thanks for digging into this @dnachev).

Not yet clear whether or not those errors require changing the way the CPs are defined or if this is a regression in ES5_GETTERS.


hjdivad:
Thanks @dnachev @rwjblue

PR #90, opened 2018-02-22
Add test for deleted records being removed from reference arrays

hjdivad:


hjdivad:
This isn't new functionality, just testing expected semantics

PR #92, opened 2018-03-02
Updated queryURL logic to use query cache for storing promise for cacheKey

dnalagatla:
Updated queryURL logic to use query cache for storing promise information for same cacheKey and upon multiple ajax invocation that promise is returned until promise is fulfilled.

### DIFF
--- a/addon/query-cache.js
+++ b/addon/query-cache.js
@@ -1,6 +1,5 @@
 import { get } from '@ember/object';
 import { A } from '@ember/array';
-import { resolve } from 'rsvp';
 
 import MegamorphicModel from './model';
 import M3RecordArray from './record-array';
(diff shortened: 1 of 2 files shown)


addon/query-cache.js
```diff
@@ -166,6 +171,11 @@ export default class QueryCache {
 
   _addResultToCache(result, cacheKey) {
     this._queryCache[cacheKey] = result;
+    //if the result is a promise
+    //skip adding to _reverseQueryCache
+    if (result instanceof Promise) {
```
dnachev: Instead of keeping two different types of data (the promise or the value) in the `_queryCache` structure, you can always keep the `Promise`. When the result has come back, the code can simply populate the reverse cache and leave the cache alone.

This should lead to simpler code, because it does not have to handle two different cases for `cachedValue`.

dnalagatla: Thank you @dnachev. It sounds good. I will take a look into this approach.

dnalagatla: I made necessary changes to clean up the logic to store the promise and update reverseCache upon success of ajax call.

addon/query-cache.js
```diff
@@ -51,18 +50,23 @@ export default class QueryCache {
...
     if (reload || cachedValue === undefined) {
+      //Store the promise in the cache until it is fulfilled and
+      //retrun same promise for subsequent request.
+      if (cacheKey) {
+        this._queryCache[cacheKey] = loadPromise;
+      }
       return loadPromise;
     } else {
-      return resolve(cachedValue);
+      return cachedValue;
```
hjdivad: Do you think it would be clearer to rename this to `cachedPromise` now that it's promises we're caching?

dnalagatla: I agree. `cachedPromise` makes more sense here now. I will update variable name.

hjdivad:
build failures are travis-ci struggling to install yarn.

PR #54, opened 2017-12-07
[Projection Support] Model data/m3 projections [0]

runspired:

PR #95, opened 2018-03-08
Projection support test only

dnachev:
This is the projection test from #54 extracted in a standalone PR and all tests skipped.

### DIFF
--- a/tests/helpers/watch-property.js
+++ b/tests/helpers/watch-property.js
@@ -1,0 +1,159 @@
+import Ember from 'ember';
+import QUnit from 'qunit';
+
+const { addObserver, removeObserver } = Ember;
+
+function makeCounter() {
+  let count = 0;
+  const counter = Object.create(null);
+  counter.reset = function resetCounter() {
+    count = 0;
+  };
+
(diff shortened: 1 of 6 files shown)


dnachev:
I'm not quite sure why the tests fails - it is complaining about the prettier format of the `projection-test.js`, but they are passing locally and reformatting the file does not yield any differences.

tests/helpers/watch-property.js
```diff
@@ -0,0 +1,113 @@
...
+      return propertyNames.reduce((result, prop) => {
+        result[prop] = watchers[prop].count;
+        return result;
+      }, {});
+    },
+    unwatch,
+  };
+}
+
+QUnit.assert.dirties = function assertDirties(options, updateMethodCallback, label) {
```
hjdivad: should probably be something like `expectDirty` no?

dnachev: This follows the QUnit `assert.throws()` pattern.

dnachev:
Closed it unintentionally.

dnachev:
OK, I've deleted the base branch, which caused this to close automatically.

PR #96, opened 2018-03-09
Use @ember/map instead of native Map

hjdivad:
ember-m3 doesn't officially support phantomjs but we won't use native
Map for now.


rwjblue:
FWIW, ember-data@3.1 will directly be using native Map...

hjdivad:
@rwjblue when it's out of canary we can drop the use of `@ember/map` here and add a polyfill in our app

Of course, moving off of phantom will be nicer still...

PR #98, opened 2018-03-09
Basic projection support

dnachev:


addon/model-data.js
```diff
@@ -44,16 +53,26 @@ export default class M3ModelData {
...
 
+    // Properties related to projections
+    this._baseModelName = this._schema.computeBaseModelName(this.modelName);
+    this._baseModelData = null;
+    this._projections = null;
+
     if (parentKey !== undefined && parentKey !== null) {
       this._parentModelData._addChildModelData(parentKey, parentValueIsArray, this);
     }
+    // TODO we may not have ID yet?
```
dnachev: This TODO is addressed in later PRs (#62).

addon/model-data.js
```diff
@@ -88,13 +113,20 @@ export default class M3ModelData {
     }
 
     if (notifyRecord) {
```
dnachev: I'm going to refactor this a little to be after `notifyProjectionProperties`, because otherwise we will have double notification for nested models (the tests for them are not enabled yet).

addon/model-data.js
```diff
@@ -138,17 +180,24 @@ export default class M3ModelData {
     this._updateChangedAttributes();
 
     if (notifyRecord) {
```
dnachev: Same - will move it after `notifyProjectionProperties`.

addon/model-data.js
```diff
@@ -318,8 +377,10 @@ export default class M3ModelData {
       }
     }
 
+    this._notifyProjectionProperties(dirtyKeys);
+
     if (notifyRecord) {
```
dnachev: Make this an `else` clause to:
`if (this._notifyProjectionProperties(dirtyKeys)) {}`

PR #99, opened 2018-03-10
Correctly notifying all projections on `set`

dnachev:
When we `set` on a projection, we delegate to the base always, but the change event was not propagated to all projections.

This PR adds this (essentially moving the `notifyPropertyChange()` call to the `setAttr()`). It has the side effect, because it will use the `storeWrapper.notifyPropertyChange()` call, which requires to be run in a runloop causing many of the existing tests to be changed slightly.

### DIFF
--- a/addon/model-data.js
+++ b/addon/model-data.js
@@ -286,6 +286,10 @@ export default class M3ModelData {
       // Add the new value to the changed attributes hash
       this._attributes[key] = value;
     }
+
+    if (!this._notifyProjectionProperties([key])) {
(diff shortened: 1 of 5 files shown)


dnachev:
The problem with the failing tests were that exceptions thrown inside an `Ember.run()` block were not re-thrown and instead the test immediately failed with the exception. Changing the nesting from:
```js
assert.throws(() => {
  run(() => {
     set(...);
  });
});
```
to 
```js
run(() => {
  assert.throws(() => {
     set(...);
  });
});
```
fixes the problem.

PR #102, opened 2018-03-21
Updated logic to keep track of unknown keys getting resolved

dnalagatla:
Updated logic to keep track of unknown keys getting resolved during resolve values.

### DIFF
--- a/addon/model-data.js
+++ b/addon/model-data.js
@@ -3,6 +3,7 @@ import { dasherize } from '@ember/string';
 import { assign, merge } from '@ember/polyfills';
 import { copy } from '@ember/object/internals';
 import SchemaManager from 'ember-m3/schema-manager';
+import { assert } from '@ember/debug';
 
 const emberAssign = assign || merge;
 
@@ -17,10 +18,47 @@ function commitDataAndNotify(modelData, updates) {
 class M3SchemaInterface {
   constructor(modelData) {
     this.modelData = modelData;
+    this._keyBeingResolved = null;
+    this._refKeyDepkeyMap = {};
+  }
+
+  _beginDependentKeyResolution(key) {
+    assert(
+      'Do not invoke `SchemaInterface` method `_beginDependentKeyResolution` without ending the resolution of previous key.',
+      this._keyBeingResolved === null
+    );
+    this._keyBeingResolved = key;
+  }
+
+  _endDependentKeyResolution(key) {
+    assert(
+      'Do not invoke `SchemaInterface` method `_endDependentKeyResolution` without begining the resolution of the key.',
+      key && this._keyBeingResolved === key
+    );
+    this._keyBeingResolved = null;
+  }
+
+  _getDependentResolvedKeys(refKey) {
+    return this._refKeyDepkeyMap[refKey];
   }
 
   getAttr(name) {
-    return this.modelData.getAttr(name);
+    let value = this.modelData.getAttr(name);
+    const keyBeingResolved = this._keyBeingResolved;
+    assert(
+      'Do not manually call methods on `schemaInterface` outside of schema resolution hooks such as `computeAttributeReference`',
+      keyBeingResolved
+    );
+
+    if (name && typeof value !== 'undefined') {
+      this._refKeyDepkeyMap[name] = this._refKeyDepkeyMap[name] || [];
+      let refKeyMap = this._refKeyDepkeyMap[name];
+      if (refKeyMap.indexOf(keyBeingResolved) < 0) {
+        refKeyMap.push(this._keyBeingResolved);
+      }
+    }
+
+    return value;
   }
 }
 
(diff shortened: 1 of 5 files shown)


addon/model.js
```diff
@@ -75,6 +75,13 @@ class EmbeddedInternalModel {
   }
 }
 
+function _computeAttributeReference(key, value, modelName, schemaInterface, schema) {
+  schemaInterface._keyBeingResolved = key;
```
dnachev: I'm generally a fan of hiding internal state manipulation behind properly named functions:
```js
schemaInterface._beginKeyResolution(key);
let reference = schema.computeAttributeReference(key, value, modelName, schemaInterface);
schemaInterface._endKeyResolution(key);
```
Later, when you need to expand the properties to notify:
```js
let dependentKeys = schemaInterface._getDependentKeys(key);
if (dependentKeys) {
  this._notifyProperties(dependentKeys);
}
```

This makes it explicitly what is the intention of `_keyBeingResolved` property is and the added complexity to implement the functionality is kept to a minimum in places of the code, which are already quite complex.

addon/model.js
```diff
@@ -100,8 +107,7 @@ function resolveValue(key, value, modelName, store, schema, model, valueInArray
     return resolvePlainArray(key, value, modelName, store, schema, model);
   }
 
-  let reference = schema.computeAttributeReference(key, value, modelName, schemaInterface);
-
+  let reference = _computeAttributeReference(key, value, modelName, schemaInterface, schema);
```
dnachev: From correctness perspective, we need to track the unknown keys for every function in `schema-manager`, which we pass the `schemaInterface` to. Right now, we really only need it for `computeAttributeReference`, because of our specific implementation of the Schema.

addon/model.js
dnalagatla: Thank you @dnachev. it's a good point and looks clean. I will update based on above suggestions.

addon/model-data.js
```diff
@@ -17,10 +18,38 @@ function commitDataAndNotify(modelData, updates) {
 class M3SchemaInterface {
   constructor(modelData) {
     this.modelData = modelData;
+    this._keyBeingResolved = null;
+    this._refKeyUnknownKeysMap = {};
+  }
+
+  _beginUnknownKeyResolution(key) {
+    assert('_keyBeingResolved should be null.', this._keyBeingResolved === null);
```
hjdivad: Can we change the assertion message to be more relevant to the user rather than focusing on internal details?

Maybe something like

```
"Do not manually call methods on `schemaInterface` outside of schema resolution hooks such as `computeAttributeReference`"
```

what do you think?

addon/model-data.js
```diff
@@ -17,10 +18,38 @@ function commitDataAndNotify(modelData, updates) {
...
+    this._keyBeingResolved = key;
+  }
+
+  _endUnknownKeyResolution(key) {
+    assert('_keyBeingResolved should be set.', key && this._keyBeingResolved === key);
```
hjdivad: same comment here as prior assert

if there's a user-error that can wind up here let's give a more user-specific message.  If we can't think of a user-error that could get them here then leaving an internal assertion message is 👍

addon/model-data.js
```diff
@@ -17,10 +18,38 @@ function commitDataAndNotify(modelData, updates) {
...
+    return this._refKeyUnknownKeysMap[refKey];
   }
 
   getAttr(name) {
-    return this.modelData.getAttr(name);
+    let value = this.modelData.getAttr(name);
+    const keyBeingResolved = this._keyBeingResolved;
+    assert('_keyBeingResolved should be set.', keyBeingResolved);
+
+    if (keyBeingResolved !== null && value) {
```
hjdivad: why do we guard on `keyBeingResolved` given the assertion on the line above?

hjdivad: Also guarding on truthiness of `value` isn't right.  We should always add `name` if it's being requested.

Imagine a flag whose value goes from `false` to `true` -- we still need to clear the cache under these conditions.

addon/model-data.js
```diff
@@ -17,10 +18,38 @@ function commitDataAndNotify(modelData, updates) {
 class M3SchemaInterface {
   constructor(modelData) {
     this.modelData = modelData;
+    this._keyBeingResolved = null;
+    this._refKeyUnknownKeysMap = {};
```
hjdivad: I'm slightly confused by the name here.

It's a dependency key map no?  why is it called `UnknownKeys`?

How about something like `_refKeyDepkeyMap`?

Or am I misunderstanding something?

addon/model-data.js
dnalagatla: Oh yeah good catch, that is the bad miss. I should check whether value is defined of not. I will update the code.

addon/model-data.js
dnalagatla: I will update, the key information.

Issue #105, opened 2018-03-22
Consistently return `RecordArray`


sangm:
In M3, since `elements` is not a tracked array, the UI does not get updated

the tl;dr is

- drop `isAttributeArrayReference`
- drop `Array.isArray` check (and `resolvePlainArray`)
- if `schema.computeAttributeReference` returns an array of references; set up a `RecordArray`

Removing this distinction makes things a bit more constrained but it should simplify things; I would expect the PR to have more deletes than additions

PR #108, opened 2018-03-24
[BUGFIX] Ensure array of references are wrapped with RecordArray

sangm:
Given an array of references, we were relying an `isAttributeArrayRefernce` to compute whether or not we  return a `RecordArray`. Clients will be confused when the UI does not get
updated even if they did not implement the API correctly.

In order to make m3 more ergonomic, we will consistently return `RecordArray`.

- Remove `isAttributeArrayReference` and let the schema handle the logic of
  returning proper data within `computeAttributeReference`.
- `computeAttributeReference` now has to take care of _either_ an array of
  references or a single reference. We will check it's type with
  `Array.isArray`, and wrap it with `RecordArray` accordingly.
- Refactored some of the methods to take in reference rather than many
  params (store, schema, etc) to ease code readability.

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -82,7 +82,7 @@ function _computeAttributeReference(key, value, modelName, schemaInterface, sche
   return reference;
 }
 
-function resolveReference(reference, store) {
+function resolveReference(store, reference) {
   if (reference.type === null) {
     // for schemas with a global id-space but multiple types, schemas may
     // report a type of null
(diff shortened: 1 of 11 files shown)


sangm:
Fixes #105

addon/model.js
```diff
@@ -94,28 +94,53 @@ function resolveReference(reference, store) {
...
+      modelName,
+      store,
+      schema,
+      schemaInterface
+    );
   }
 
+  if (Array.isArray(value)) {
+    return resolvePlainArray(key, value, modelName, store, schema, model);
+  }
```
hjdivad: can we drop this `isArray(value)` check and instead have `schema.computeNestedModel` return an array (ie let userland determine whether this is an array of nested models or not?

If that's at all a pain i'm 👍  on doing that in a separate pr as well

sangm: I'll definitely work on this.

I'll fix this with a follow up PR

PR #109, opened 2018-03-29
Projection support for embedded models

dnachev:
This is rebased and updated version of the #65 PR.

Couple of notes:
* Had to refactor how child model datas are created
* There is one missing test case for handling deletions because of updates.

### DIFF
--- a/addon/model-data.js
+++ b/addon/model-data.js
@@ -72,16 +72,7 @@ class M3SchemaInterface {
 }
 
 export default class M3ModelData {
-  constructor(
-    modelName,
-    id,
-    clientId,
-    storeWrapper,
-    parentModelData,
-    parentKey,
-    parentValueIsArray,
-    embeddedInternalModel
-  ) {
+  constructor(modelName, id, clientId, storeWrapper, parentModelData, baseModelData) {
     this.modelName = modelName;
     this.clientId = clientId;
     this.id = id;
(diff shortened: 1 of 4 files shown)


addon/model-data.js
```diff
@@ -517,6 +540,14 @@ export default class M3ModelData {
       // destroy
       delete this._childModelDatas[key];
     }
+    if (this._projections) {
+      // TODO Add a test for this destruction
+      for (let i = 0; i < this._projections.length; i++) {
```
dnachev: We can start from `i = 1` here, as we know that the base is always the very first element. This will save us the `if` statement below.

addon/model-data.js
```diff
@@ -696,9 +727,10 @@ export default class M3ModelData {
   }
 
   _notifyRecordProperties(changedKeys) {
-    if (this._embeddedInternalModel) {
+    if (this._embeddedInternalModel && this._embeddedInternalModel.record) {
```
hjdivad: Why the conjunction?

When is it valid to have an embedded internal model with no record?

addon/model-data.js
```diff
@@ -696,9 +727,10 @@ export default class M3ModelData {
...
       this._embeddedInternalModel.record._notifyProperties(changedKeys);
-    } else {
+    } else if (!this._parentModelData) {
+      // only notify through the store if it is not a child model data
```
hjdivad: if there is no embedded internal model this cannot be an embedded (child) model data.

let's keep these conditions simpler and instead add asserts

addon/model-data.js
```diff
@@ -93,25 +84,19 @@ export default class M3ModelData {
...
-    this._baseModelData = null;
+    this._baseModelData = baseModelData;
     this._projections = null;
 
-    if (parentKey !== undefined && parentKey !== null) {
-      this._parentModelData._addChildModelData(parentKey, parentValueIsArray, this);
-    }
-
-    if (this._baseModelName) {
+    if (this._baseModelName || this._baseModelData) {
```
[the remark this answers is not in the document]
dnachev: This can happen, when a nested property has been projected as its base model, e.g. the `address` field of a company is projected as `Address` instead of `CountryOnlyAddress`.

addon/model-data.js
dnachev: When we do `pushData` for the embedded model, it will attempt to notify the records, but `record` field of `_embeddedInternalModel` is assigned only after `pushData` has been called. If we re-order these calls, we can get rid of the additional check for record on the embedded internal model.

addon/model-data.js
dnachev: The flow is the following:
- a `Company` projection named `MiniCompany`  with `address` field projected as `CountryOnlyAddress`.
- when we read `address` from the projection, it will create two model datas:
  - one in the context of `MiniCompany` model data, with a type `CountryOnlyAddress`
  - one in the context of `Company` model data, with type `Address`.
 However, we only receive an embedded internal model for the `CountryOnlyAddress` model data, because the `get()` is done for the projection only. We will populate the `embeddedInternalModel` of the base `Address` model data, when the code calls `get('address')` on the base `Company` model.

dnachev:
Address some of the comments and left replies to the rest.

PR #121, opened 2018-04-12
Added support for validation errors in a record.

dnalagatla:
- Added unit tests for validation errors support in model

### DIFF
--- a/tests/unit/model-test.js
+++ b/tests/unit/model-test.js
@@ -4010,4 +4010,113 @@ module('unit/model', function(hooks) {
       true
     );
   });
+
+  test('errors is intialized upon creation of the record', function(assert) {
+    let model = run(() =>
+      this.store.push({
+        data: {
+          id: 'isbn:9780439708180',
+          type: 'com.example.bookstore.Book',
+          attributes: {
+            name: `Harry Potter and the Sorcerer's Stone`,
+          },
+        },
+        included: [],
+      })
+    );
+
+    // Testing errors is getting set and is an instance
+    // of DS.Errors
+    const errors = get(model, 'errors');
+    assert.ok(errors);
+    assert.ok(errors instanceof DS.Errors);
+  });
+
+  test('.save errors getting updated via the store and removed upon setting a new value', function(assert) {
+    assert.expect(10);
+
+    this.owner.register(
+      'adapter:-ember-m3',
+      EmberObject.extend({
+        updateRecord(store, type, snapshot) {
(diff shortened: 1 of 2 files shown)


tests/unit/model-test.js
```diff
@@ -4010,4 +4010,108 @@ module('unit/model', function(hooks) {
...
+
+    assert.equal(model.get('isSaving'), false, 'initially model not saving');
+
+    run(() => {
+      model.set('estimatedPubDate', '_invalidEstimatedPublishDate');
+    });
+
+    // Testing client validation errors getting
+    // set upon getting validation errors from API
+    run(() => {
```
hjdivad: let's return this promise chain.

It turns out not to matter right now but that's only because of an implementation detail in ember-testing that is a bit odd.

This test code *looks* like it has a race condition and I'd rather have it be more standard even if it's not strictly necessary.

So something more or less like

```js
return run(() =>
  // no {} so implicit return
  model.save().catch(() => {
    // ...
  }).then(() => {
    // Testing remove errors upon setting a new value
   // ...
  })
);
```

tests/unit/model-test.js
```diff
@@ -4010,4 +4010,108 @@ module('unit/model', function(hooks) {
...
+          get(error, 'firstObject.message'),
+          'Please enter valid estimated publish date',
+          'error is available'
+        );
+      });
+    });
+
+    //Testing remove errors upon setting a new value
+    run(() => set(model, 'estimatedPubDate', 'January 2621'));
+    assert.equal(get(model, 'errors.length'), 1, 'error is removed upon setting new value');
```
hjdivad: let's add the following

 - assert the model is still invalid
 - set the rest of the fields
 - assert the model is now valid

tests/unit/model-test.js
```diff
@@ -4010,4 +4010,113 @@ module('unit/model', function(hooks) {
...
+    assert.equal(model.get('isSaving'), false, 'initially model not saving');
+
+    run(() => {
+      model.set('estimatedPubDate', '_invalidEstimatedPublishDate');
+    });
+
+    // Testing client validation errors getting
+    // set upon getting validation errors from API
+    return run(() => {
+      model.save().catch(() => {
```
hjdivad: we need to return the promise here; the example i posted earlier returned it implicitly because it was `() => X` instead of `() => { X }`

anyway we just need `return model.save(). // ...`

dnalagatla: Updated the code to return the promise

PR #129, opened 2018-05-08
Fix `changedAttributes` for projections

dnachev:
The `changedAttributes()` did not delegate to the base model data, when appropriate leading to a bug where:
```javascript
projection.set('property', newValue);
projection.changedAttributes() // == {}
```

Open question: ideally, we should filter out the `changedAttributes()` result based on the attribute whitelist from the schema, the question is where it is more appropriate to do - in the `model-data` or in the `model`? I think it is easier to be done in `model-data` as the code already iterates over the nested model datas.

### DIFF
--- a/addon/model-data.js
+++ b/addon/model-data.js
@@ -350,6 +350,9 @@ export default class M3ModelData {
     @private
   */
   changedAttributes() {
+    if (this._baseModelData) {
+      return this._baseModelData.changedAttributes();
+    }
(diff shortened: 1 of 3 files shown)


hjdivad:
@dnachev `model-data` seems like the right place to me.  Is there any reason to prefer `model`?

dnachev:
Only because this is where the schema is mostly consulted, nothing technical.

Issue #127, opened 2018-05-04
add schema hook for setting values


hjdivad:
the hook should intercept `modelData.setAttr`

it should provide a schema interface for setting the attr, which is the default behaviour.

this would allow someone to intercept a set to a resolved value and instead update the model data with an unresolved value.  in either case the resolved value is cached (as per normal) on the model itself.

at least one use case for this is multiple projections that resolve the same data differently (eg to different types).

hjdivad:
fixed by #133

PR #133, opened 2018-05-29
Added schema hook for setAttr #127

dnalagatla:


tests/unit/projection-test.js
```diff
@@ -1780,6 +1780,257 @@ module('unit/projection', function(hooks) {
...
+      });
+
+      this.records = {
+        baseRecord,
+        projectedExcerpt,
+        projectedPreview,
+      };
+
+      //Adding .setAttribute hook in schema
+      SchemaManager.schema.setAttribute = function(modelName, attr, value, schemaInterface) {
```
hjdivad: shouldn't we do this before we push anything into the store?

tests/unit/projection-test.js
```diff
@@ -1780,6 +1780,257 @@ module('unit/projection', function(hooks) {
     });
   });
 
+  module('Update projection property with resolved value', function(hooks) {
+    // properties for use for initial state
+    const BOOK_ID = 'isbn:9780439708181';
+    // TODO is this valid? we won't have a real ID yeah?
```
hjdivad: not following this TODO; what's the issue?

hjdivad:
@dnalagatla can we update `README.md` with `setAttribute` as well?

hjdivad:
- [ ] as you observed, we should also remove [this code](https://github.com/hjdivad/ember-m3/blob/40c03c55fc769dcf7bc867d55ae951fa680978cd/addon/model.js#L524-L538) and replace it with a call to the `setAttribute` hook

tests/unit/projection-test.js
dnalagatla: Oh, I copied projection update from a another test module, missed removing the TODO. I will fix that.

tests/unit/projection-test.js
dnalagatla: That is a good point.

But Looking at the code of `store.push`, it doesn't invoke `.set` method on the model, as result probably I didn't see any error. 

But I will update the code to set the hook before invoking the store.push.

Issue #134, opened 2018-05-30
Add mutation support for arrays of nested models


hjdivad:
- [ ] when `schema.computeNestedModel` indicates an array property is an array of nested models, don't return a plain array; return a custom array similar to, but not actually, `addon/record-array`

- [ ] the nested model array accepts POJOs in `replace` (and therefore in `unshiftObject`, `pushObject` &c.)

- [ ] these pojos get converted into nested models via `schema.computeNestedAttribute`; it's an error to `pushObject` a pojo the schema doesn't understand on a `NestedModelArray`

---

- [ ] both `NestedModelArray#replace` and `RecordArray#replace` can update their underlying `modelData` so that the new values appear in `changedAttributes()`

hjdivad:
@dnalagatla can we close this?  you've landed it no?

PR #136, opened 2018-06-01
Added support to track changes in array #134

dnalagatla:


tests/unit/model-test.js
```diff
@@ -3207,6 +3210,129 @@ module('unit/model', function(hooks) {
...
+
+    // Pushing simple value to array
+    run(() => {
+      chapters.pushObject({
+        name: 'Windy eh new',
+        number: 1,
+      });
+    });
+
+    assert.ok(chapters.get('lastObject') instanceof MegamorphicModel);
```
hjdivad: let's add an assertion that checks the new model has the right attrs

```js
assert.equal(chapters.get('lastObject').get('name'), 'Windy eh new', 'new embedded model has right attrs');
```

addon/model.js
```diff
@@ -167,21 +163,45 @@ function resolveValue(key, value, modelName, store, schema, model, parentIdx) {
...
+  if (value && value.length > 0) {
+    resolvedArray = value.map((value, idx) =>
+      resolveValue(key, value, modelName, store, schema, model, idx)
+    );
   }
 
-  return value.map((value, idx) => resolveValue(key, value, modelName, store, schema, model, idx));
+  return M3PlainArray.create({
+    content: A(resolvedArray),
+    _replace(idx, removeAmt, newItems) {
```
hjdivad: I'm not a fan of passing in `_replace` here; I think it obscures the way the arrays function too much.

`resolve` can be imported by the various record arrays: i'd rather we create the arrays with the relevant context that they stash during `init` than pass in a function that has that context in a closure.

addon/model.js
```diff
@@ -167,21 +163,45 @@ function resolveValue(key, value, modelName, store, schema, model, parentIdx) {
...
-function resolveRecordArray(store, references) {
+function resolveRecordArray(store, model, key, references) {
   let recordArrayManager = store._recordArrayManager;
 
   let array = M3RecordArray.create({
     modelName: '-ember-m3',
     content: A(),
     store: store,
     manager: recordArrayManager,
+    _replace(idx, removeAmt, newModels) {
```
hjdivad: as with the array for nested models, it'll be clearer if we pass in context rather than a `_replace` function;

something like

```js
M3RecordArray.create({
  /* existing fields ... */
 model,
});
```

addon/record-array.js
```diff
@@ -6,6 +6,11 @@ export default class extends RecordArray {
   // TODO: implement more of RecordArray but make this not an arrayproxy
 
   replace(idx, removeAmt, newModels) {
+    if (typeof this._replace === 'function') {
```
hjdivad: instead of doing this kind of check let's split this

`RecordArray` - base class
`QueryArray < RecordArray` - used by `queryURL`
`ReferenceArray < RecordArray` - used for arrays of attribute references

addon/model.js
```diff
@@ -167,21 +163,45 @@ function resolveValue(key, value, modelName, store, schema, model, parentIdx) {
 
 // ie an array of nested models
 function resolvePlainArray(key, value, modelName, store, schema, model) {
```
hjdivad: instead of relying on `resolvePlainArray` for this let's allow `schema.computeNestedModel` to return an array.  This makes the API simpler and more consistent with how we handle `computeAttributeReference`

So if `computeNestedModel` returns an array we create an m3 nested array and if it returns `undefined` (as well as `computeAttributeReference` we'll fall back to a plain array of *unresolved* values

what do you think?

addon/m3-plain-array.js
```diff
@@ -0,0 +1,16 @@
+import ArrayProxy from '@ember/array/proxy';
```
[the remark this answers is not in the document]
dnalagatla: how about  `M3TrackedArray` basically store any values which are not references.

tests/unit/model-test.js
dnalagatla: Added assertions for the attributes. And also fixed test failures locally.

addon/model.js
```diff
@@ -322,7 +198,7 @@ export default class MegamorphicModel extends EmberObject {
     let oldValue = this._cache[key];
     let newValue = this._internalModel._modelData.getAttr(key);
 
-    let oldIsRecordArray = oldValue && oldValue instanceof M3RecordArray;
+    let oldIsRecordArray = oldValue && oldValue instanceof M3ReferenceArray;
```
hjdivad: is there a reason to check against `M3ReferenceArray` instead of `M3RecordArray`?

I think we might want `M3RecordArray` as this will still catch reference arrays, but it will also keep the same functionality for people who `model.set(attr, resultOfQueryURL)`

addon/model.js
```diff
@@ -499,7 +375,7 @@ export default class MegamorphicModel extends EmberObject {
 
     if (isArray(value)) {
       const cachedValue = this._cache[key];
-      if (cachedValue instanceof M3RecordArray) {
+      if (cachedValue instanceof M3ReferenceArray) {
```
hjdivad: as before i think we want s/M3ReferenceArray/M3RecordArray here

addon/model.js
dnalagatla: oh ok. I will fix that.

PR #137, opened 2018-06-07
Fix `computeNestedModel` access to schemaInterface

dnachev:
Despite the fact we were passing `schemaInterface` to the `computeNestedModel`, it wasn't correctly set, which caused an exception to be thrown.

Added a test for it as well.


dnalagatla:
LGTM

PR #138, opened 2018-06-14
Associate newly added records to record arrays

hjdivad:
`RecordArray`s created for either `queryURL` or reference arrays
register themselves with `internalModel`s so that they'll automatically
update when those models are unloaded.

Now, register with models that are added to these arrays so that they
also share this functionality.


dnalagatla:
looks good to me.

PR #142, opened 2018-06-25
Convert schema manager to be a service

dnachev:
As discussed:
- Because `SchemaManager` is a global, any schema implementations, which have to retain state run the risk of state sharing between separate Fastboot requests. Problems, caused by this state sharing, will be extremely hard to debug as they are timing based.
- Converting the `SchemaManager` to a service allows applications to reset the state of their schema as they see fit.
- I've kept the `registerSchema` API to allow for both stateless and stateful schema implementations.

The alternatives I've considered:
- Require the schema implementation to be available as a service, which the `SchemaManager` will inject or lookup as needed.
- Require the application to extend `SchemaManager` base class and provide the actual service, which the rest of M3 will use. Dismissed, because it makes evolving `SchemaManager` harder (it has to consider it's subclasses) without much gain.
- Completely remove the implementation of `SchemaManager` and require applications to implement all functions (right now, there are a couple, which simply lookup the `models` property of the schema). I find this appealing, but I also don't know why we've started with the current setup, so I'm dismissing it for now.


README.md
```diff
@@ -57,13 +57,14 @@ For example, if your API returns responses like the following:
...
-// app/initializers/schema-initializer.js
+// app/instance-initializers/schema-initializer.js
 const BookstoreRegExp = /^com\.example\.bookstore\.*/;
 const ISBNRegExp = /^isbn:.*/;
 const URNRegExp = /^urn:(\w+):(.*)/;
 
-export function initialize() {
-  SchemaManager.registerSchema({
+export function initialize(application) {
+  let schemaManager = application.lookup('m3-schema-manager');
```
hjdivad: `service:m3-schema-manager`

PR #143, opened 2018-06-26
Update ember data

dnalagatla:


addon/model-data.js
```diff
@@ -183,6 +183,10 @@ export default class M3ModelData {
 
   removeFromHasMany() {}
 
+  _initRecordCreateOptions(options) {
```
dnachev: Why is this method starting with an underscore? It seems part of the public Record Data API as it is not internal to ember-m3.

runspired: because we need to RFC it in ember-data and it's going to have to evolve a bit still

PR #147, opened 2018-07-03
Cache url

### LINKED ISSUE #140: add api for inserting directly into the query cache
```js
store.queryURL(url, { cacheKey }).then(result => {
  store.cacheURL( computeSecondaryCacheKey(cacheKey, result), result);
})
```

unsure about name, intended to fit with [these](https://github.com/hjdivad/ember-m3/blob/master/addon/initializers/m3-store.js#L57-L67)

---

This is helpful for eg querying an seo-friendly "clean" url but still caching it for a subsequent app generated "dirty" url.

sangm:
Initial work for cacheURL. Fixes #140 

Few things to consider:

- What do we do with "reverseCache" _if_ anything?

### DIFF
--- a/addon/query-cache.js
+++ b/addon/query-cache.js
@@ -1,3 +1,5 @@
+import { Promise as RSVPPromise } from 'rsvp';
+import { assert } from '@ember/debug';
 import { get } from '@ember/object';
 import { A } from '@ember/array';
 
@@ -68,6 +70,16 @@ export default class QueryCache {
     } else {
       return cachedPromise;
     }
+  }
+
+  cacheURL(cacheKey, result) {
+    assert(
+      'cacheKey must be a non-empty string',
+      typeof cacheKey === 'string' && cacheKey.length > 0
+    );
+    // We will wrap `result` as a promise to be type consistent.
+    this._queryCache[cacheKey] = RSVPPromise.resolve(result);
+    this._addResultToReverseCache(result, cacheKey);
   }
 
   unloadRecord(record) {
--- a/tests/unit/query-cache-test.js
+++ b/tests/unit/query-cache-test.js
@@ -922,4 +922,54 @@ module('unit/query-cache', function(hooks) {
       assert.equal(this.adapterAjax.called, false, '`adapter.ajax` is not called');
     });
   });
+
+  test('.cacheURL inserts consistently inserts a promise into cache', function(assert) {
+    assert.expect(3);
+    const mockResult = { id: 1, foo: 'bar' };
+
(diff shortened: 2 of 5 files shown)


addon/query-cache.js
```diff
@@ -70,6 +71,12 @@ export default class QueryCache {
     }
   }
 
+  cacheURL(cacheKey, result) {
+    if (cacheKey) {
```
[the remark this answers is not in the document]
sangm: Assertion sounds good

addon/query-cache.js
```diff
@@ -70,6 +72,13 @@ export default class QueryCache {
     }
   }
 
+  cacheURL(cacheKey, result) {
+    assert('cacheKey cannot be falsey', cacheKey);
```
hjdivad: maybe `cacheKey must be a non-empty string`?

tests/unit/query-cache-test.js
```diff
@@ -922,4 +922,52 @@ module('unit/query-cache', function(hooks) {
...
+
+    return cachedResult.then(result => {
+      assert.deepEqual(mockResult, result, 'queryURL returned cached result');
+    });
+  });
+
+  test('.cacheURL supports unloading of models and invalidates cache', function(assert) {
+    assert.expect(2);
+
+    const firstPayload = {
```
hjdivad: I prefer `let` for locally-scoped vars like this unless we're actually `Object.freeze`ing

that said I don't care much, but we should at least be consistent between `firstPayload` and `secondPayload` as neither appears to be reassigned.

tests/unit/query-cache-test.js
```diff
@@ -922,4 +922,52 @@ module('unit/query-cache', function(hooks) {
...
+    assert.expect(2);
+    const mockResult = { id: 1, foo: 'bar' };
+
+    this.queryCache.cacheURL('foo-bar', mockResult);
+    const cachedResult = this.queryCache.queryURL('/foo-bar', { cacheKey: 'foo-bar' });
+
+    assert.ok(cachedResult && cachedResult.then, 'cachedResult is a promise');
+
+    return cachedResult.then(result => {
+      assert.deepEqual(mockResult, result, 'queryURL returned cached result');
```
hjdivad: can we add an assertion that `this.adapterAjax` was not called (similar to the `.queryURL caches its results when given a cacheKey` test)

sangm:
Fixed the issues you mentioned 👍

hjdivad:
awesome thanks @sangm 

If you want to add something to `README.md` that would be great but otherwise no worries i'll write something up next time i release

PR #157, opened 2018-07-19
[DOCS] Adding documentation for `cacheURL `

sangm:


sangm:
As requested @hjdivad  😄

PR #160, opened 2018-07-20
API documentation

xg-wang:
This PR take the first step to add jsdoc to code, also fix some legacy API description in the readme.


addon/initializers/m3-store.js
```diff
@@ -56,6 +59,18 @@ export function extendStore(Store) {
       return this._super(modelName, id, clientId, storeWrapper);
     },
 
+    /**
+     * A thin wrapper around the API response that knows how to look up relationships.
```
hjdivad: I'd be tempted to change `relationships` to `references`.  In practice they serve similar functionality but they have different semantics so I'd rather use a different term in an attempt to avoid confusing people.

addon/initializers/m3-store.js
```diff
@@ -56,6 +59,18 @@ export function extendStore(Store) {
...
+     *
+     * @param {string} url The URL path to query.
+     * @param {Object} options
+     * @param {string} [options.method=GET] The HTTP method to use.
+     * @param {Object} [options.params] The parameters to include
+     * @param {string} [options.cacheKey] A string to uniquely identify this request.
+     * @param {boolean} [options.reload=false]
```
hjdivad: "If true, block on a new request even if a cached value exists"
maybe?

addon/initializers/m3-store.js
```diff
@@ -56,6 +59,18 @@ export function extendStore(Store) {
...
+     * @param {boolean} [options.backgroundReload=false]
```
hjdivad: "If true and a cached value exists, issue a non-blocking request but immediately fulfill with the cached value"

addon/initializers/m3-store.js
```diff
@@ -56,6 +59,18 @@ export function extendStore(Store) {
...
+     * @returns {Promise}
```
hjdivad: It would be great to add the type of the resolution value as well.

maybe something like
```js
@returns {Promise<M3Model|RecordArray,Error>} a promise for loading `url` that fulfills to an `M3Model` if the response is a single resource or a `RecordArray` of `M3Model`s if the response is an array of resources.
```

addon/model-data.js
```diff
@@ -51,6 +68,12 @@ class M3SchemaInterface {
     return this._refKeyDepkeyMap[refKey];
   }
 
+  /**
+   * Get the attribute of the model data plus recording the key being resolved
```
hjdivad: Since this is public API I think it makes more sense to talk about server payloads than model data, which is a concept for addon authors and internals, but not really for app developers.

Maybe something like

"Get the attribute `name` from the model.  This can be useful if your payload keys are different from your attribute names; for example if your api adds a prefix to attributes that should be interpreted as references."

addon/schema-manager.js
```diff
@@ -3,18 +3,58 @@ export class SchemaManager {
...
   }
 
-  computeAttributeReference(key, value, modelname, schemaInterface) {
-    return this.schema.computeAttributeReference(key, value, modelname, schemaInterface);
+  /**
+   * Determines whether an attribute is a reference.
+   * If it is not, return `null` or `undefined`.
+   * Otherwise return an object with properties:
+   *  - `id` The id of the referenced model (either m3 or `DS.Model`)
+   *  - `type` The type of the referenced model (either m3 or `DS.Model`)
```
hjdivad: `null` is also a valid type in which case `id` will be looked up in a global cache

hjdivad:
This looks great; thanks @xg-wang.  Couple minor suggestions inline.

README.md
```diff
@@ -81,11 +83,11 @@ export function initialize(application) {
           id: value,
           type: 'com.example.bookstore.Book',
         };
-      } else if (match = URNRegExp.exec(value)) {
+      } else if ((match = URNRegExp.exec(value))) {
```
hjdivad: prettier's overdoing it a bit with these parens ^_^

hjdivad:
This is great, thanks @xg-wang!

Issue #162, opened 2018-07-24
ensure childModelData index is in sync with trackedarray index


hjdivad:
this index is wrong: https://github.com/hjdivad/ember-m3/blob/master/addon/m3-tracked-array.js#L35

- [ ] also consider only removes
- [ ] also consider state of items to the right

cc @dnalagatla

hjdivad:
@dnalagatla this is fixed by #162 isn't it?

PR #166, opened 2018-07-31
ensure childModelData index is in sync with trackedarray index #162

dnalagatla:


addon/m3-tracked-array.js
```diff
@@ -25,20 +25,35 @@ export default class M3TrackedArray extends ArrayProxy {
   }
 
   replaceContent(idx, removeAmt, newItems) {
-    newItems = newItems.map((item, idx) => {
+    // Update childModelDatas array
+    // mapping to array of nested models
+    const childModelDatas = this._model._internalModel._modelData.__childModelDatas;
+    const key = this._key;
+    if (childModelDatas && childModelDatas[key] && Array.isArray(childModelDatas[key])) {
```
hjdivad: This array check should probably be an assertion

addon/m3-tracked-array.js
```diff
@@ -25,20 +25,35 @@ export default class M3TrackedArray extends ArrayProxy {
...
+      const newItemsInChildModelData = new Array(newItems.length);
+      Array.prototype.splice.apply(
+        childModelDatas[key],
+        [idx, removeAmt].concat(newItemsInChildModelData)
+      );
+    }
+
+    let startIdx = idx;
+    newItems = newItems.map(item => {
       if (isResolvedValue(item)) {
```
hjdivad: I think we should merge this pr without resolving this issue as it's an edge case that we don't actually handle

BUT

Can we please log a bug to remind ourselves to test this case; in particular, to test that inserting a resolved value into a tracked array (and especially at some position other than the start or end) leaves everything in a correct state

addon/m3-tracked-array.js
```diff
@@ -25,20 +25,35 @@ export default class M3TrackedArray extends ArrayProxy {
...
+    // mapping to array of nested models
+    const childModelDatas = this._model._internalModel._modelData.__childModelDatas;
+    const key = this._key;
+    if (childModelDatas && childModelDatas[key] && Array.isArray(childModelDatas[key])) {
+      const newItemsInChildModelData = new Array(newItems.length);
+      Array.prototype.splice.apply(
+        childModelDatas[key],
+        [idx, removeAmt].concat(newItemsInChildModelData)
+      );
+    }
```
hjdivad: I think it's confusing to have tracked array know so much about the internals of modeldata.

I suspect there's something better we can do, but for now let's do something like add `modelData._resizeChildModelData(idx,removeLength,addLength)`  so model data can own the logic here rather than tracking array know about its internals

addon/m3-tracked-array.js
```diff
@@ -25,20 +25,35 @@ export default class M3TrackedArray extends ArrayProxy {
...
+
+    let startIdx = idx;
+    newItems = newItems.map(item => {
       if (isResolvedValue(item)) {
         associateRecordWithRecordArray(item, this);
         return item;
       }
 
+      const index = startIdx;
+      startIdx++;
```
hjdivad: i would personally see this more clearly if the index were computed as the sum of `idx` and the index argument to `map`

addon/m3-tracked-array.js
dnalagatla: Added an assertion.

dnalagatla: Added an assertion.

addon/m3-tracked-array.js
dnalagatla: Moved the implementation to `modelData._resizeChildModelData`

addon/m3-tracked-array.js
dnalagatla: updated the code to compute the index.

PR #169, opened 2018-08-06
Schema as a service

hjdivad:
Instead of registering the schema via a global it is now looked up at `service:m3-schema` and can therefore participate in the rest of the container infrastructure.

Among other things, this makes it easier to manage state in schemas, as well as have app schemas depend on other services.

---

Also add support for tracked arrays to add embedded models (including
from other top models).  This is a weird use case and not recommended,
but there's no particular reason for it not to work.


xg-wang:
Since schema is a service, does it make sense to remove `m3-schema-manager` and use schema service directly? 
This way user only needs to customize their `m3-schema` service and call methods like  `getDefaultValue` from the service. Rather than injecting another `m3-schema-manager` or duplicating code in app's `m3-schema`.

hjdivad:
@xg-wang yes I definitely think that's an improvement.

I would still want a level of indirection; but that level of indirection doesn't need to be via a service.  All I really want is for there to not be any methods in the [default schema](https://github.com/hjdivad/ember-m3/blob/master/addon/services/m3-schema.js) besides the public API.

PR #183, opened 2018-09-05
Remove exporting m3-schema

xg-wang:
User should extend DefaultSchema and export from app/services. Plus fix some sinon deprecation.


hjdivad:
thanks @xg-wang

PR #199, opened 2018-10-09
[CHORE recordData] rename modelData to recordData

runspired:
As the RFC for `ModelData` finalized, it was determined that `RecordData` would be a more appropriate feature name, and `ember-data` recently update the implementation to reflect this detail.

Here, I've done so as well. For ensuring the widest compatibility possible with existing users of `M3` and for existing versions of `ember-data` that `M3` users might be using, three things have been done:

1) accessing `_recordData` via `internalModel` or `record._internalModel` is now done via a util function `recordDataFor` that abstracts whether `_modelData` or `_recordData` is in use.
2) We use `IS_RECORD_DATA` from `ember-compatibility-helpers` to target `recordData` versions of things that are `3.5+` while using `modelData` versions of things for `<= 3.5-beta`


addon/initializers/m3-store.js
```diff
@@ -51,11 +51,23 @@ export function extendStore(Store) {
...
+    createModelDataFor() {
+      let ret = this.createRecordDataFor(...arguments);
+
+      // fallback to older ember-data methodName super if needed
+      if (ret === undefined) {
+        return this._super(...arguments);
+      }
+
+      return ret;
+    },
```
hjdivad: I'm not a fan of adding this complexity at runtime.

What are the versions involved here?

`createModelDataFor` is 3.4.x and `createRecordDataFor` is 3.5.0 right?

can we split these up with https://github.com/ember-cli/ember-cli-version-checker ?

addon/m3-reference-array.js
```diff
@@ -11,7 +11,7 @@ export default class M3ReferenceArray extends M3RecordArray {
   init() {
     super.init(...arguments);
     this._key = get(this, 'key');
-    this._model = get(this, 'model');
+    this._record = get(this, 'model');
```
hjdivad: rename arg to `record` as well?

addon/-private.js
```diff
@@ -0,0 +1,5 @@
+export function recordDataFor(recordOrInternalModel) {
+  let internalModel = recordOrInternalModel._internalModel || recordOrInternalModel;
+
+  return internalModel._recordData || internalModel._modelData;
+}
```
hjdivad: can we do this at build time with  https://github.com/ember-cli/ember-cli-version-checker  instead of adding this complexity at runtime?

if we have to do it at runtime, add comments to make it clear what code belongs to what version of ember data (3.4.x and 3.5.0 are the relevant versions i think)

addon/record-data.js
```diff
@@ -0,0 +1,989 @@
...
+    baseRecordData
+  ) {
+    this.modelName = modelName;
+    this.clientId = clientId;
+    this.id = id;
+    this.storeWrapper = storeWrapper;
+
+    if (!storeWrapper.recordDataFor) {
+      storeWrapper.recordDataFor = storeWrapper.modelDataFor;
+    }
```
hjdivad: rather than mutate the `storeWrapper` we get from ember-data let's wrap access to this in a helper (similar to `recordDataFor` and swap out its implementation at build time based on the version of ember data

addon/record-data.js
```diff
@@ -0,0 +1,989 @@
...
+    this.__childRecordDatas = null;
+    this._schema = schemaManager;
+
+    this.schemaInterface = new M3SchemaInterface(this);
+
+    // Properties related to projections
+    this._baseRecordData = baseRecordData;
+    this._projections = null;
+
+    this._initBaseModelData();
```
hjdivad: rename to record data?

addon/record-data.js
```diff
@@ -0,0 +1,989 @@
...
+      this.__inFlightAttributes = Object.create(null);
+    }
+    return this.__inFlightAttributes;
+  }
+
+  set _inFlightAttributes(v) {
+    this.__inFlightAttributes = v;
+  }
+
+  _initBaseModelData() {
```
hjdivad: rename to record data?

runspired: renamed

addon/record-data.js
runspired: renamed

addon/record-data.js
runspired: done

addon/m3-reference-array.js
runspired: renamed

addon/initializers/m3-store.js
runspired: done

addon/-private.js
runspired: done

runspired:
@hjdivad all updates complete

PR #215, opened 2018-11-07
Fixing nested model updates from local updates and partial updates from server.

ygongdev:
Was able to reproduce a scenario where if you change a nested model and called save(), the changed attributes actually persist instead of being reset, so this new test will fail.

Investigation: Inside `willCommit`, we have `childRecordData.willCommit()` which never got called because childRecordData is an array of 2 elements, ~~which I think holds the old and new record.However, I'm not sure if the isArray check there is intentional for this scenario.~~ `DidCommit` only gets called once for the top-level model and the `_mergeUpdates()` call never happens.

Theoretically the test name above this new test was a bit misleading because it didn't actually test the changedAttribute() behavior at all.

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -211,6 +211,7 @@ export default class M3RecordData {
     if (this._baseRecordData) {
       return this._baseRecordData.willCommit();
     }
+    // debugger
     this._inFlightAttributes = this._attributes;
     this._attributes = null;
 
@@ -220,9 +221,9 @@ export default class M3RecordData {
         let childKey = nestedKeys[i];
         let childRecordData = this._childRecordDatas[childKey];
         if (!Array.isArray(childRecordData)) {
-          // we don't re-use nested models within arrays so there's no need to
-          // propagate willCommit/didCommit
           childRecordData.willCommit();
+        } else {
+          childRecordData.forEach(child => child.willCommit());
         }
       }
     }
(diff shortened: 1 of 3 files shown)


ygongdev:
If I understand correctly, according to the documentation for `ember-data`,
`_data` holds all of the attributes that have been acknowledged by a backend via the adapter``

In this scenario where we change the nested model. After the `save` promise resolves, e.g `save().then(resolvedModel => ...)`,
`resolvedModel._internalModel._recordData._data` should contain the changes made on its nested model. But that doesn't seem to be the case because I kept seeing the old `_data`.
The `childRecordData`, `resolvedModel._internalModel._recordData._childRecordData.[key][index]` contains the changes in its `_attributes`. 
To me, it seems like that currently if we make a change to a nested model, the backend is not acknowledging the changes properly.

ygongdev:
If I understand correctly, according to the documentation for ember-data,
`_data holds all of the attributes that have been acknowledged by a backend via the adapter`

In this scenario where we change the nested model. After the save promise resolves, e.g `save().then(resolvedModel => ...)`,
`resolvedModel._internalModel._recordData._data` should contain the changes made on its nested model. But that doesn't seem to be the case because I kept seeing the old _data.
The `childRecordData`, e.g `resolvedModel._internalModel._recordData._childRecordData.[key][index]` contains the changes in its `_attributes`.
To me, it seems like that currently if we make a change to a nested model, the backend is not acknowledging the changes properly.

ygongdev:
If I understand correctly, according to the documentation for ember-data,
_data holds all of the attributes that have been acknowledged by a backend via the adapter

In this scenario where we change the nested model. After the save promise resolves, e.g save().then(resolvedModel => ...),
resolvedModel._internalModel._recordData._data should contain the changes made on its nested model. But that doesn't seem to be the case because I kept seeing the old _data.
The childRecordData, e.g resolvedModel._internalModel._recordData._childRecordData.[key][index] contains the changes in its _attributes.
To me, it seems like that currently if we make a change to a nested model, the backend is not acknowledging the changes properly.

ygongdev:
Explanation for the changes.
`_syncNestedModelUpdates` recursively calls `didCommit` on its children, so their `_data` is in sync with the changes.
This is important because previously the `sync` could've only happened when we received a response from the server. Because there are potentially local changes within nested models that the server response didn't account for, therefore, we should call `didCommit` on these changes.

`_mergeNestedModelData()` basically put the children `_data` into its parent's `_data`. The reason is that we want the top level model's `_data` to be up-to-date.
By the time we called `_mergeNestedModelData()` inside `didCommit`, the updates from the server and/or local would be done, so we can be sure that when we merge, we have the correct `_data`.
`_mergeNestedModelData()` is not inside `_mergeUpdates` purely because `_mergeUpdates` is also called on `pushData`, so it'll call `_mergeNestedModelData()`, which doesn't seem to make sense at that point in time and it throws some weird error.

I also added 6 tests for the different scenarios where the updates could happen:
1. Server response with nested non-array models should override local changes.
2. Server response with nested array models should override local changes.
3. Local changes on nested non-array models without server response should update the model correctly.
4. Local changes on nested array models without server response should update the model correctly.
5. Partial updates from the server response and local updates on nested non-array models.
6. Partial updates from the server response and local updates on nested array models.

igorT:
This looks pretty 👍 to me. I rebased your changes as https://github.com/hjdivad/ember-m3/tree/igor/rebase-215, so you might want to replace this branch with that one

ygongdev:
@igorT Sure, how do I go about that? Sorry not too familiar to the workflow of the github UI and git CLI combo.

hjdivad:
looks good thanks @ygongdev @igorT

PR #223, opened 2018-11-08
Fixes rollbackAttributes for record data from projections

eddie-ruva:
Calling rollbackAttributes on a record data that has a base record data
was not working properly because the attributes mutated were in
the record data from the projection and not the base record data.

When dealing with projections all the interactions should be delegated
to the base record data.

In this PR I am adding a fix to call rollback attributes on the base
record data when one is present.

While adding a test for this I noticed that a similar bug was present in `isAttrDirty` which I've also addressed.

### DIFF
--- a/tests/unit/record-data-test.js
+++ b/tests/unit/record-data-test.js
@@ -198,6 +198,29 @@ module('unit/record-data', function(hooks) {
     assert.equal(rollbackAttributesSpy.getCalls().length, 0, 'rollbackAttributes was not called');
   });
 
+  test('.rollbackAttributes rolls backs the attributes on a base record data when dealing with a projection', function(assert) {
+    assert.expect(1);
+    const projectedRecordData = this.storeWrapper.recordDataFor(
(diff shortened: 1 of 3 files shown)


tests/unit/record-data-test.js
```diff
@@ -198,6 +198,29 @@ module('unit/record-data', function(hooks) {
...
+      recordDataKey({
+        modelName: 'com.bookstore.book',
+        id: '1',
+      })
+    ];
+    const rollbackAttributesSpy = this.sinon.spy(baseRecordData, 'rollbackAttributes');
+
+    projectedRecordData.rollbackAttributes();
+
+    assert.equal(
```
igorT: Can you add a test for `isAttrDirty` here please?

eddie-ruva: @igorT I've added a test for `isAttrDirty` here. I thought about removing the one in `record-data` but wanted to get your thoughts on that before doing so. 

Let me know if you'd like to see any other changes.

PR #232, opened 2018-11-21
Fix issues in deep copy

hjdivad:
1. cycle detection was broken
2. it's not a generic copy utility: we don't need or want to copy
  non-JSON values which will sometimes appear in `attributes`.

It's perfectly fine for such values to be in `attributes` as they will
likely have serialization paths in users' serializers (eg if they are
models), but there is no need to copy such values when calculating
changed attributes.

### DIFF
--- a/addon/utils/copy.js
+++ b/addon/utils/copy.js
@@ -8,24 +8,30 @@ function _copy(obj, seen, copies) {
 
   // avoid cyclical loops
   // eslint-disable-next-line no-cond-assign
-  if ((loc = seen.indexOf(obj) >= 0)) {
+  if ((loc = seen.indexOf(obj)) >= 0) {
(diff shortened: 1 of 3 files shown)


addon/utils/copy.js
```diff
@@ -8,24 +8,30 @@ function _copy(obj, seen, copies) {
...
-  // object will need to implement Copyable.
   if (Array.isArray(obj)) {
     ret = obj.slice();
+    copies.push(ret);
 
     loc = ret.length;
 
     while (--loc >= 0) {
-      ret[loc] = copy(ret[loc], copies);
+      ret[loc] = _copy(ret[loc], copies);
```
rwjblue: This seems wrong, should be passing seen

PR #236, opened 2018-12-12
Fix changedAttributes for newly created nested records

igorT:
Previously if you started with a parent record looking like:
`{ id: 1, name: 'David' }` and set a new nested property on it:
`record.set('nestedKey', { prop: 'hi' })`, `record.changedAttributes()`
would return `{ nestedKey: { prop: [undefined, 'hi'] } }`

This makes it impossible to detect a newly created nested record.
This commit changes the return value if the record is newly created to:
would return `{ nestedKey: [ undefined, { prop: [undefined, 'hi'] } ] }`


addon/record-data.js
```diff
@@ -421,6 +421,17 @@ export default class M3RecordData {
...
   }
+  /**
+   * @param {string} key
+   * @returns {boolean}
+   */
+  getServerAttr(key) {
+    if (this._baseRecordData) {
+      return this._baseRecordData.getServerAttr(key);
+    } else {
+      return this._data[key];
```
eddie-ruva: I think we might need to check explicitly for `null` otherwise boolean attributes that are set to false would be treated as non existent in the server. What do you think?

igorT: Nice catch

igorT: Will add this check to the call sites

PR #237, opened 2018-12-12
Fix the build import of notifyPropertyChange

igorT:


runspired:
@hjdivad we don't need this and did confirm that, we may need to update the test harness for the other things.

The gist was that the shim is provided by ember-cli-babel.

hjdivad:
@runspired mmm the pr was failing due to 2.18 lacking the API.  Can you put up a pr that uses the shim instead?

PR #241, opened 2019-01-03
Add `isAttributeResolved` schema hook

hjdivad:
When attributes are set, they will be immediately cached if they are
resolved, but go through normal resolution when first accessed
otherwise.

This hook gives users control over whether or not a newly set value is
considered resolved.

It also fixes a bug where empty plain arrays were treated as resolved.
By default, arrays of records are resolved and non-native arrays (eg
`RecordArray`s, `ManyArray`s &c.) are treated as resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,6 +399,24 @@ have following properties.
   }
   ```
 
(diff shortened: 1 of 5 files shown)


igorT:
👍

PR #259, opened 2019-02-06
Upgrade deps

hjdivad:


hjdivad:
main motivation is to get rid of `ember-rvc176-data@0.3.5` which causes problems when linking

PR #260, opened 2019-02-06
[BUGFIX ReferenceArray.replace] fix error when replace is called without prototype extensions

runspired:
also adds test for adding and then deleting a projection from a resolution array attached to a newly created parent projection


hjdivad:
LGTM although in future let's split that test file up rather than keep around this 3k test file

PR #266, opened 2019-02-21
Hjdivad/batch change notfiications

hjdivad:


addon/model.js
```diff
@@ -390,7 +398,8 @@ export default class MegamorphicModel extends EmberObject {
         // We update record arrays in-place to match the semantics of setting a
         // `hasMany` attribute on a `DS.Model`.
         this._setRecordArray(key, value);
-        notifyPropertyChange(this, key);
+        this._deferProprtyChange(key);
+        flushChanges(this._store);
```
igorT: Why do we defer and then immediately flush?

addon/model.js
```diff
@@ -413,6 +422,7 @@ export default class MegamorphicModel extends EmberObject {
 
     // Remove errors upon setting of new value
     this._removeError(key);
+    flushChanges(this._store);
```
igorT: Why do we have to flush here if we are not in the push but rather in the `set` flow?

addon/model.js
```diff
@@ -226,10 +221,14 @@ export default class MegamorphicModel extends EmberObject {
       // TODO: disconnect recordData -> childRecordData in the case of nested model -> primitive
       // anything -> undefined | primitive
       delete this._cache[key];
-      super.notifyPropertyChange(key);
+      this._deferProprtyChange(key);
     }
   }
 
+  _deferProprtyChange(key) {
+    deferPropertyChange(this._store, this, key);
```
igorT: It seems like defer logic is only called here, I'm curious about the tradeoffs of having the defer logic in the util method vs putting it on the store. It seems that we create a weak map to map stores to changes, but to me it seems natural for the store to track/manage this?

addon/model.js
```diff
@@ -290,12 +289,17 @@ export default class MegamorphicModel extends EmberObject {
   }
 
   rollbackAttributes() {
+    if (DEBUG) {
+      assertNoChanges(this._store);
```
igorT: I like this.

addon/model.js
```diff
@@ -290,12 +289,17 @@ export default class MegamorphicModel extends EmberObject {
...
+    }
+
     let dirtyKeys = recordDataFor(this).rollbackAttributes();
     this._updateCurrentState(loadedSaved);
 
     if (dirtyKeys && dirtyKeys.length > 0) {
       this._notifyProperties(dirtyKeys);
     }
+    flushChanges(this._store);
```
igorT: If we start with 0 changes, and don't push, why do we have to flush?

addon/record-data.js
```diff
@@ -174,14 +174,24 @@ export default class M3RecordData {
   }
 
   /**
-   * @param {Object} jsonApiResource
-   * @param {boolean} calculateChange
+   * Notify this `RecordData` that it has new attributes from the server.
+   *
+   * @param {Object} jsonApiResource the payload resource to use for updating
+   * the server attributes
+   * @param {boolean} calculateChange Whether or not changes that result from
```
igorT: Missing the new arg in the docstring

addon/record-data.js
```diff
@@ -207,7 +217,14 @@ export default class M3RecordData {
...
 
-    if (this._notifyProjectionProperties(changedKeys)) {
+    // by default, always notify projections when we receive data.  We might
+    // not have been asked to calculate changes if the base record data has
+    // no record, but we might still have records instantiated for our
+    // projections.
+    //
+    // Notifications are explicitly suppressed when we're using `pushData`
+    // synthetically while resolving nested record data
+    if (!suppressNotifications && this._notifyProjectionProperties(changedKeys)) {
```
igorT: Is `suppressNotifications` actually `suppressProjectionNotification`? Otherwise, we dont't honor it on `232`
We seem to be increasing the number of notification flags, wondering if there is a way to refactor away?

addon/initializers/m3-store.js
```diff
@@ -87,6 +88,16 @@ const STORE_OVERRIDES = {
     return this._queryCache.contains(cacheKey);
   },
 
+  // override _push to batch change notifications which we're obliged to do
+  // since all our properties are treated as volatiles as they come from
+  // `unknownProperty`
+  _push(jsonApiDoc) {
+    beginChanges(this);
```
igorT: Is `beginChanges` unnecessary given that `deferChanges` does that?

addon/utils/notify-changes.js
```diff
@@ -0,0 +1,53 @@
...
+if (HasNotifyPropertyChange) {
+  notifyPropertyChange = _notifyPropertyChange;
+} else {
+  notifyPropertyChange = propertyDidChange;
+}
+
+const StoreToChanges = new WeakMap();
+
+export function beginChanges(store) {
+  StoreToChanges.set(store, []);
```
igorT: Whats the advantage of this approach vs keeping track of this on the store?

addon/utils/notify-changes.js
```diff
@@ -0,0 +1,53 @@
...
+    for (let i = 0; i < changes.length; i += 2) {
+      let obj = changes[i];
+      let key = changes[i + 1];
+      notifyPropertyChange(obj, key);
+    }
+  });
+  StoreToChanges.set(store, []);
+}
+
+export function assertNoChanges(store) {
```
igorT: I really like this, is there a way to maybe schedule it at the end of a runloop every time we call `defer` so we are always sure to not have stale notifications?

addon/utils/notify-changes.js
```diff
@@ -0,0 +1,53 @@
...
+    beginChanges(store);
+  }
+  let changes = StoreToChanges.get(store);
+  changes.push(obj, key);
+}
+
+export function flushChanges(store) {
+  let changes = StoreToChanges.get(store) || [];
+  changeProperties(() => {
+    for (let i = 0; i < changes.length; i += 2) {
```
igorT: I assume we know this data layout is best for perf?

addon/model.js
hjdivad: No reason; i'll change it to be explicitly immediate to better signify intent

addon/model.js
hjdivad: the change notification gets deferred in [RecordData.setAttr](https://github.com/hjdivad/ember-m3/blob/hjdivad/batch-change-notfiications/addon/record-data.js#L387)

It seems clearer to me to have a flush here rather than have the change notification be immediate deeper in the system (in record data) but i'm open to being persuaded otherwise

addon/model.js
hjdivad: as we discussed OOB; `WeakMap` is used for private state only; conceptually it is managed by the store

addon/model.js
hjdivad: yeah in general we could use a more aggressive assertions in the codebase

addon/model.js
hjdivad: because `_notifyProperties` goes through the same codepath that defers change notifications

hjdivad: see [notifyPropertyChange](https://github.com/hjdivad/ember-m3/blob/c94be430117e7b19dac6ab33018ea609fe7935a9/addon/model.js#L193-L226)

addon/record-data.js
hjdivad: good call; i'll rename the flag

re: flags I very much agree; i'd like to see if we can clean this up in a followup pr

addon/utils/notify-changes.js
hjdivad: not exposing private api to users

addon/utils/notify-changes.js
hjdivad: that's a great idea.  #269

addon/utils/notify-changes.js
hjdivad: what's the layout you're referring to specifically?

addon/initializers/m3-store.js
hjdivad: fair enough; i meant for it to better signal intent, but i'll drop it

PR #270, opened 2019-03-06
chore: add tests for didSave

hjdivad:
Currently "batching" only works for payloads that contain included
resources.  In principle it is easy to resolve if we flush in
`didSaveRecord` for payloads that do not contain included resources, but
there is no way for us to detect this and there isn't a method we can
override from ember data as this behaviour exists in a function within a
closure.


hjdivad:
see #271

PR #273, opened 2019-03-14
feat: enable computeNestedModel to handle arrays

runspired:
Previously if an array was not a reference array we did not give users a chance to handle it in-bulk vs by individual member. This makes extraction of some forms of schema information that needs to be either index-aware or aware that it is within a collection of some form difficult to detect. This change allows users to handle.

BREAKING: This change will break users who currently check for `value` in `computeNestedModel` being of type `object` but do not check whether it is an `Array`. The refactor is tiny, with the simplest refactor being to return `null` directly to maintain previous behavior.

e.g.

```js
computeNestedModel(key, value) {
  if (Array.isArray(value)) { return null; }
  // continue on ..
}
```

### DIFF
--- a/addon/resolve-attribute-util.js
+++ b/addon/resolve-attribute-util.js
@@ -113,6 +113,11 @@ function transferOrResolveValue(store, schema, record, recordData, modelName, ke
 }
 
 function createNestedModel(store, record, recordData, key, nestedValue, parentIdx = null) {
+  if (parentIdx !== null && nestedValue instanceof EmbeddedMegamorphicModel) {
+    recordData._setChildRecordData(key, parentIdx, recordDataFor(nestedValue));
+    return nestedValue;
+  }
+
   let internalModel = new EmbeddedInternalModel({
     // nested models with ids is pretty misleading; all they really ought to need is type
     id: nestedValue.id,
@@ -151,10 +156,6 @@ function createNestedModel(store, record, recordData, key, nestedValue, parentId
       nestedRecordData.setAttr(key, nestedValue.attributes[key], true);
     });
   }
-
-  // if (parentIdx !== null) {
-  //   recordData._setChildRecordData(key, parentIdx, nestedRecordData);
-  // }
 
   return nestedModel;
 }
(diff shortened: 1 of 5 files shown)


addon/resolve-attribute-util.js
```diff
@@ -95,11 +95,13 @@ export function resolveValue(key, value, modelName, store, schema, record, paren
...
   }
 
-  if (Array.isArray(value)) {
-    return resolveArray(key, value, modelName, store, schema, record, recordData);
-  }
   let nested = computeNestedModel(key, value, modelName, schemaInterface, schema);
   if (nested) {
+    // handle the case where nested returns an array.
+    if (Array.isArray(nested)) {
+      return resolveArray(key, value, modelName, store, schema, record, recordData);
```
hjdivad: This doesn't look right to me.  If `computeNestedModel` returns an array, it should be similar to what it returns in the non-array case, eg

```js
[{
  id: 'some-id-or-null',
  type: 'some-type',
  attributes: { / * ... */ },
}, {
  id: 'some-id-or-null',
  type: 'some-type',
  attributes: { / * ... */ },
}]
```

but your call to `resolveArray` is passing `value`.  It's not clear to me how this would work.

hjdivad: Also we need tests for this case, where `computeNestedModel` returns an array and app code is able to resolve it correctly

runspired: added

tests/unit/model/changed-attrs-test.js
```diff
@@ -1275,6 +1275,91 @@ module('unit/model/changed-attrs', function(hooks) {
...
+          if (value !== null && typeof value === 'object') {
+            return { id: key, type: attributesType, attributes: value };
+          }
+        }
+
+        computeBaseModelName(modelName) {
+          return ['com.bookstore.projected-book', 'com.bookstore.excerpt-book'].includes(modelName)
+            ? 'com.bookstore.book'
+            : null;
+        }
```
hjdivad: can we delete `computeBaseModelName` here?  Is the test actually using projections?

PR #274, opened 2019-03-26
Feat: kill of Model dependence on EmberObject

runspired:
The deprecation noise was annoying (and slows down test runs a lot). I notice it when working with projects consuming ember-m3 and with ember-m3 itself.

We should do some more testing before merging that we don't have any other weird reliances on `EmberObject` inheritance, and that we make sure we do the right work in `destroy` (if there are things to do).

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -384,10 +384,6 @@ export default class MegamorphicModel {
       return;
     }
 
-    if (!this._init) {
-      console.log({ key, value });
-    }
-
     if (DEBUG) {
       assertNoChanges(this._store);
     }
--- a/addon/model.js
+++ b/addon/model.js
@@ -3,7 +3,8 @@
 
 import DS from 'ember-data';
 import { RootState } from 'ember-data/-private';
-import EmberObject, { computed, get, set, defineProperty } from '@ember/object';
+import { computed, get, set, setProperties, defineProperty } from '@ember/object';
+import { addObserver } from '@ember/object/observers';
 import { isArray } from '@ember/array';
 import { assert, warn } from '@ember/debug';
 import { readOnly } from '@ember/object/computed';
@@ -102,41 +103,50 @@ const retrieveFromCurrentState = computed('_topModel.currentState', function(key
   return this._topModel._internalModel.currentState[key];
 }).readOnly();
 
-// global buffer for initial properties to work around
-//  a)  can't write to `this` before `super`
-//  b)  core_object writes properties before calling `init`; this means that no
-//      CP or setknownProperty can rely on any initialization
-let initProperites = Object.create(null);
+const INIT_KEYS = {
+  store: 1,
+  _internalModel: 1,
+  _topModel: 1,
+  _parentModel: 1,
+  isError: 1,
+};
 
-export default class MegamorphicModel extends EmberObject {
+export default class MegamorphicModel {
+  constructor(properties) {
+    this.init(properties);
+  }
   init(properties) {
-    // Drop Ember.Object subclassing instead
-    super.init(...arguments);
+    this._init = false;
+    this.store = properties.store;
     this._store = properties.store;
     this._internalModel = properties._internalModel;
     this._cache = Object.create(null);
     this._schema = get(properties.store, '_schemaManager');
 
-    this._topModel = this._topModel || this;
-    this._parentModel = this._parentModel || null;
-    this._errors = new DS.Errors();
-    this._init = true;
+    this._topModel = properties._topModel || this;
+    this._parentModel = properties._parentModel || null;
+    this._errors = DS.Errors.create();
+    this.isDestroying = false;
+    this.isDestroyed = false;
+    this.isError = properties.isError;
 
-    this._flushInitProperties();
-  }
-
-  _flushInitProperties() {
-    let propertiesToFlush = initProperites;
-    initProperites = Object.create(null);
-
-    let keys = Object.keys(propertiesToFlush);
+    let keys = Object.keys(properties);
     if (keys.length > 0) {
       for (let i = 0; i < keys.length; ++i) {
         let key = keys[i];
-        let value = propertiesToFlush[key];
-        this.setUnknownProperty(key, value);
+        let value = properties[key];
+        if (INIT_KEYS[key] !== 1) {
+          set(this, key, value);
+        }
       }
     }
+
+    this._init = true;
+  }
+
+  destroy() {
+    this.isDestroying = true;
+    this.isDestroyed = true;
   }
 
   static get isModel() {
@@ -180,6 +190,10 @@ export default class MegamorphicModel extends EmberObject {
     // currentState is defined on the prototype and will be treated as
     // non-volatile, so it's safe to eagerly send a change event
     notifyPropertyChange(this, 'currentState');
+  }
+
+  get(key) {
+    return get(this, key);
   }
 
   notifyPropertyChange(key) {
@@ -247,6 +261,14 @@ export default class MegamorphicModel extends EmberObject {
 
   set(key, value) {
     set(this, key, value);
+  }
+
+  setProperties(properties) {
+    setProperties(this, properties);
+  }
+
+  addObserver(...args) {
+    addObserver(this, ...args);
   }
 
   serialize(options) {
@@ -357,13 +379,13 @@ export default class MegamorphicModel extends EmberObject {
   // TODO: drop change events for unretrieved properties
   setUnknownProperty(key, value) {
     if (key === OWNER_KEY) {
+      this[OWNER_KEY] = value;
       // 2.12 support; later versions avoid this call entirely
       return;
     }
 
     if (!this._init) {
-      initProperites[key] = value;
-      return;
+      console.log({ key, value });
     }
 
     if (DEBUG) {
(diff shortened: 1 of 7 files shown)


addon/model.js
```diff
@@ -102,41 +103,50 @@ const retrieveFromCurrentState = computed('_topModel.currentState', function(key
   return this._topModel._internalModel.currentState[key];
 }).readOnly();
 
-// global buffer for initial properties to work around
-//  a)  can't write to `this` before `super`
-//  b)  core_object writes properties before calling `init`; this means that no
-//      CP or setknownProperty can rely on any initialization
-let initProperites = Object.create(null);
+const INIT_KEYS = {
```
hjdivad: Please don't use all caps for var names.

Kind of annoying that we can't configure eslint for this.  🤷‍♂️ 

`InitKeys` or `Init_Keys` is fine.

addon/model.js
```diff
@@ -102,41 +103,50 @@ const retrieveFromCurrentState = computed('_topModel.currentState', function(key
...
+  store: 1,
+  _internalModel: 1,
+  _topModel: 1,
+  _parentModel: 1,
+  isError: 1,
+};
```
hjdivad: why are we using `1` here instead of `true`?  it looks like we're treating this as booleans

addon/model.js
```diff
@@ -102,41 +103,50 @@ const retrieveFromCurrentState = computed('_topModel.currentState', function(key
...
-    let keys = Object.keys(propertiesToFlush);
+    let keys = Object.keys(properties);
     if (keys.length > 0) {
       for (let i = 0; i < keys.length; ++i) {
         let key = keys[i];
-        let value = propertiesToFlush[key];
-        this.setUnknownProperty(key, value);
+        let value = properties[key];
+        if (INIT_KEYS[key] !== 1) {
+          set(this, key, value);
```
hjdivad: why do we need to `set` these during `init`?

addon/model.js
```diff
@@ -102,41 +103,50 @@ const retrieveFromCurrentState = computed('_topModel.currentState', function(key
...
+        }
       }
     }
+
+    this._init = true;
```
hjdivad: is this used anywhere now that `flushInitProperties` is gone?

addon/model.js
```diff
@@ -249,6 +263,14 @@ export default class MegamorphicModel extends EmberObject {
     set(this, key, value);
   }
 
+  setProperties(properties) {
+    setProperties(this, properties);
+  }
+
+  addObserver(...args) {
+    addObserver(this, ...args);
+  }
```
hjdivad: Are we sure we need this?

If we do to avoid practical breakage then fair enough but i'm surprised at this one

addon/model.js
```diff
@@ -357,15 +379,11 @@ export default class MegamorphicModel extends EmberObject {
   // TODO: drop change events for unretrieved properties
   setUnknownProperty(key, value) {
     if (key === OWNER_KEY) {
+      this[OWNER_KEY] = value;
       // 2.12 support; later versions avoid this call entirely
```
hjdivad: ah we can drop this soon once 3.9 releases

tests/unit/query-array-test.js
```diff
@@ -17,7 +17,7 @@ module('unit/query-array', function(hooks) {
     }();
 
     this.createRecordArray = function(options = {}) {
-      return new M3QueryArray(
+      return M3QueryArray.create(
```
hjdivad: Wouldn't it be better to drop `ArrayProxy` as the base class?

tests/unit/query-array-test.js
```diff
@@ -60,7 +60,7 @@ module('unit/query-array', function(hooks) {
   });
 
   test('QueryArray requires a query', function(assert) {
-    let queryArray = new M3QueryArray();
+    let queryArray = M3QueryArray.create();
```
hjdivad: Wouldn't it be better to drop `ArrayProxy` as the base class?

tests/unit/record-array-test.js
```diff
@@ -40,7 +40,7 @@ module('unit/record-array', function(hooks) {
     });
 
     this.createRecordArray = function() {
-      let recordArray = new M3RecordArray();
+      let recordArray = M3RecordArray.create();
```
hjdivad: Wouldn't it be better to drop `ArrayProxy` as the base class?

rwjblue:
Hmm, I'm somewhat worried about the impact of this (e.g. making it harder for us to land in apps).

I did a minimal thing to be able to run the tests without any deprecations in case that route might be preferred:

https://github.com/hjdivad/ember-m3/compare/master...rwjblue:prevent-deprecations-without-removing-base-class

^ emits a clean console (no warnings / deprecations at all) and is _not_ a breaking change to consumers...

addon/model.js
stefanpenner: ya, seems boolean is a more natural fit

addon/model.js
stefanpenner: We should aim to drop, and not support.

tests/unit/query-array-test.js
stefanpenner: yes please (if possible)

addon/model.js
```diff
@@ -102,41 +103,50 @@ const retrieveFromCurrentState = computed('_topModel.currentState', function(key
...
+  _parentModel: 1,
+  isError: 1,
+};
 
-export default class MegamorphicModel extends EmberObject {
+export default class MegamorphicModel {
+  constructor(properties) {
+    this.init(properties);
+  }
   init(properties) {
```
stefanpenner: do we ever reinvoke `init` ? if not, we may want to inline.

addon/model.js
```diff
@@ -102,41 +103,50 @@ const retrieveFromCurrentState = computed('_topModel.currentState', function(key
...
-
-    this._flushInitProperties();
-  }
-
-  _flushInitProperties() {
-    let propertiesToFlush = initProperites;
-    initProperites = Object.create(null);
+    this._topModel = properties._topModel || this;
+    this._parentModel = properties._parentModel || null;
+    this._errors = DS.Errors.create();
```
stefanpenner: can we lazily create this?

addon/model.js
```diff
@@ -102,41 +103,50 @@ const retrieveFromCurrentState = computed('_topModel.currentState', function(key
...
+    this.init(properties);
+  }
   init(properties) {
-    // Drop Ember.Object subclassing instead
-    super.init(...arguments);
+    this._init = false;
+    this.store = properties.store;
     this._store = properties.store;
     this._internalModel = properties._internalModel;
     this._cache = Object.create(null);
```
stefanpenner: Can we lazily alloc? (assuming it isn't simply immediately accessed.

addon/model.js
```diff
@@ -102,41 +103,50 @@ const retrieveFromCurrentState = computed('_topModel.currentState', function(key
...
     this._schema = get(properties.store, '_schemaManager');
```
stefanpenner: At what version of ember can we simply do `properties.store._schemaManager` ?

hjdivad:
@rwjblue can you post it as a PR?  That one seems unambiguously worth merging now.

I still like @runspired's approach of removing the base class but we can land your minimal changes more easily to kill off the deprecation warnings.

addon/model.js
stefanpenner: looks to be used when setting id during init: https://github.com/hjdivad/ember-m3/blob/66a28db3998b03af27ace254f738c05af22b70c8/addon/model.js#L363

It could be restructured, by putting `id` in `INIT_KEYS` and assigning manually to the module during construction. Might make sense to do so? Seems kinda strange to keep around the `_init` flag only for a reason that could be avoided.

stefanpenner:
Awesome, ya this should get off `Ember.Object` asap.

addon/model.js
hjdivad: Ah yeah IIRC we can't trivially work around this in ember-m3 b/c of the way ember-data handles initial id

addon/model.js
hjdivad: @stefanpenner yep that's what i was getting at 👍

addon/model.js
runspired: We don't, but we do stub it in tests currently

addon/model.js
runspired: Yes, and we should move to that for sure

addon/model.js
runspired: _cache is accessed nearly immediately

addon/model.js
runspired: `set` delegates to the prototype / setters first if something is defined, and falls back to `setUnknownProperty` for anything remaining. This thereby ensures we mirror our current behavior when letting EmberObject's super call handle the initial prop setting.

addon/model.js
runspired: We had one test using it in M3, happy to refactor that test, was just trying to feel out the surface area of what was getting used.

tests/unit/query-array-test.js
runspired: Yes, but that might take more work. Will investigate when I get a chance :)

addon/model.js
hjdivad: We know which are which though don't we?

can't we `set` the ones that need to go through `setUnknownProperty` and do a regular property assignment for those that don't need special treatment?

rwjblue:
@hjdivad - https://github.com/hjdivad/ember-m3/pull/276

rwjblue:
Pairing with @hjdivad, closing this for now but we should totally come back and make this happen when we have the time.

PR #276, opened 2019-03-28
Prevent deprecations without changing base class

rwjblue:


hjdivad:
Merging since the only failure is 2.18 which we'll drop support for in next release as it's out of LTS maint.

Issue #280, opened 2019-04-26
Do not Batch errors


hjdivad:
#92 introduced batching for #63.  This prevents us from making two requests when a second one is made before the first completes.

However, it introduced a regression in caching semantics, as errors (promise rejections from `adapter.ajax(/*...*/)`) are cached instead of only fulfillment values.

In the success case, keeping the cache a promise cache (rather than result cache) is transparent to the user, but in the error case we should evict the cache when the request completes.

hjdivad:
fixed in #282

PR #282, opened 2019-04-26
Evict Promise cache for rejected responses to queryURL

teopalva:
This PR addresses the issue described in https://github.com/hjdivad/ember-m3/issues/280, where promises fulfilling with an error response are still cached by queryURL. This results in the undesired behavior that subsequent identical requests after a first failure are automatically rejected, thus preventing retries.


hjdivad:
thanks @teopalva!

PR #298, opened 2019-06-27
feat: batch array changes

hjdivad:
Previously only property changes were batched; now array changes in
record arrays are also batched.  Note that this is not the case for
tracked arrays, but this feature should be added there as well.

As part of this change, remove `ArrayProxy` as the base class of
`RecordArray`.  Although our arrays are array proxies, the base class is
not adding value and it's easier to implement all of the logic
ourselves.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 **/blueprints/schema/files/app/services/m3-schema.js
+**/node_modules/**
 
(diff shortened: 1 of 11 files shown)


hjdivad:
Testing in-app for additional context.

Also this is a breaking change.  Although I don't expect it to actually impact apps.

PR #354, opened 2019-08-16
fix: base RecordData creation when creating a projection

runspired:
M3 was relying on an unknown implicit behavior to create new client side records without using createRecord. This broke when `clientId` began to be unique across types. This change prepares M3 for versions of EmberData where `clientId` is unique across types when paired with https://github.com/emberjs/data/pull/6334

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -4,7 +4,7 @@ import { assign, merge } from '@ember/polyfills';
 import { copy } from './utils/copy';
 import { assert } from '@ember/debug';
 import Ember from 'ember';
-import { IS_RECORD_DATA } from 'ember-compatibility-helpers';
+import { IS_RECORD_DATA, gte } from 'ember-compatibility-helpers';
 
 const emberAssign = assign || merge;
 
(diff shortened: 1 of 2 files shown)


addon/record-data.js
```diff
@@ -696,11 +696,15 @@ export default class M3RecordData {
       }
 
       if (IS_RECORD_DATA) {
-        this._baseRecordData = this.storeWrapper.recordDataFor(
-          dasherize(baseModelName),
-          this.id,
-          this.clientId
-        );
+        if (gte('ember-data', '3.13.0')) {
+          this._baseRecordData = this.storeWrapper.recordDataFor(dasherize(baseModelName), this.id);
```
igorT: shouldn't this be 3.13.1 or 3.14?

runspired: No because 3.13 is still beta

hjdivad:
Do we expect this to resolve issues on beta? https://travis-ci.org/hjdivad/ember-m3/jobs/573846095

runspired:
We do but while the commit has been backported there is not yet a beta cut with it

hjdivad:
@runspired got it; thx

PR #363, opened 2019-08-21
fix: a couple of bugs with references

hjdivad:
1. Allow user schemas to compute references with `null` ids and given
  types (which will result in `null` results)
2. Prioritize the user schema's `computeAttributeReference` over the raw
  value when deciding whether to build a reference array.

### DIFF
--- a/addon/resolve-attribute-util.js
+++ b/addon/resolve-attribute-util.js
@@ -12,19 +12,20 @@ import {
 } from './utils/resolve';
 
 function resolveReference(store, reference) {
+  let { id } = reference;
   if (reference.type === null) {
     // for schemas with a global id-space but multiple types, schemas may
     // report a type of null
-    let internalModel = store._globalM3Cache[reference.id];
+    let internalModel = store._globalM3Cache[id];
     return internalModel ? internalModel.getRecord() : null;
   } else {
     // respect the user schema's type if provided
-    return store.peekRecord(reference.type, reference.id);
+    return id !== null && id !== undefined ? store.peekRecord(reference.type, reference.id) : null;
   }
 }
 
 function resolveReferenceOrReferences(store, model, key, value, reference) {
-  if (Array.isArray(value) || Array.isArray(reference)) {
+  if (Array.isArray(reference)) {
     return resolveRecordArray(store, model, key, reference);
   }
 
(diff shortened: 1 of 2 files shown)


hjdivad:
cc @rwjblue

addon/resolve-attribute-util.js
```diff
@@ -12,19 +12,20 @@ import {
...
   } else {
     // respect the user schema's type if provided
-    return store.peekRecord(reference.type, reference.id);
+    return id !== null && id !== undefined ? store.peekRecord(reference.type, reference.id) : null;
   }
 }
 
 function resolveReferenceOrReferences(store, model, key, value, reference) {
-  if (Array.isArray(value) || Array.isArray(reference)) {
+  if (Array.isArray(reference)) {
```
runspired: what's the motivation here? Isn't that at odds with the concept of "always sync"?

addon/resolve-attribute-util.js
```diff
@@ -12,19 +12,20 @@ import {
...
   if (reference.type === null) {
     // for schemas with a global id-space but multiple types, schemas may
     // report a type of null
-    let internalModel = store._globalM3Cache[reference.id];
+    let internalModel = store._globalM3Cache[id];
     return internalModel ? internalModel.getRecord() : null;
   } else {
     // respect the user schema's type if provided
-    return store.peekRecord(reference.type, reference.id);
+    return id !== null && id !== undefined ? store.peekRecord(reference.type, reference.id) : null;
```
runspired: when do we have this case of having a reference but not a record?

hjdivad: The algorithm of check for ref then nested model then raw value isn’t great; we should change it to one hook on the user schema where they pick (this saves us from some silly dancing between computeAttributeReference and computeNestedModel).

Cases where you want to explicitly say i have a reference with no record are mainly where you have a difference between the object properties the user accesses and the underlying keys in the payload

addon/resolve-attribute-util.js
hjdivad: Broadly it's to always prioritize user signal over convention or heuristics.

Specifically if you have a payload like

```js
{
  foreignKeys: ['urn:book:1'],
  resolvedForeignKeys: null // due to resolution failure; perhaps b/c a user deleted the record or something
}
```

Whether this should be an empty reference array or `null` should be up to the user

PR #370, opened 2019-08-26
fix: stringify deeply nested data and minor refactoring in debug-adapter

SYU15:
Here are a couple of quick follow-up improvements for `debug-adapter`. Not sure if the test is what you had in mind based on your feedback (seems my test is causing me to stub out more, not less, out of necessity), but happy to refactor/re-write as needed.

### DIFF
--- a/addon/adapters/debug-adapter.js
+++ b/addon/adapters/debug-adapter.js
@@ -1,6 +1,6 @@
 import { A, isArray } from '@ember/array';
 import DataAdapter from '@ember/debug/data-adapter';
-import { get, setProperties, defineProperty } from '@ember/object';
+import { get, defineProperty } from '@ember/object';
 import { inject } from '@ember/service';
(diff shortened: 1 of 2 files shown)


SYU15:
Since it was a relatively minor change, have also rolled in the fix for handling deeply nested data to be rendered in Ember Inspector.

Unfortunately, for now, we need to stringify objects/arrays that are more than 2 levels deep since this is explicitly not handled by Ember Inspector: https://github.com/emberjs/ember-inspector/blob/ab6fead7b64173d0fc78443290df65cfe538ebc4/ember_debug/object-inspector.js#L71. 

Maybe we should fix it there and remove this code? But for now, this will enable developers to get a more granular idea of what their data is.

**Before**
Screen Shot 2019-08-29 at 5 44 02 PM
**After**
Screen Shot 2019-08-29 at 5 43 03 PM

hjdivad:
thanks @SYU15 !

PR #375, opened 2019-08-29
fix: arrayDidChange called with correct args

hjdivad:
Previously `replace` did not pass the index to array observers
correctly.

Also enable `setupTestIsolationValidation`.

Co-authored-by: Robert Jackson <rjackson@linkedin.com>

### DIFF
--- a/tests/test-helper.js
+++ b/tests/test-helper.js
@@ -4,4 +4,6 @@ import { start } from 'ember-qunit';
 import './helpers/watch-property';
 
 setResolver(resolver);
-start();
+start({
+  setupTestIsolationValidation: true,
+});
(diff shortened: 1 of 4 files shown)


tests/test-helper.js
```diff
@@ -4,4 +4,6 @@ import { start } from 'ember-qunit';
 import './helpers/watch-property';
 
 setResolver(resolver);
-start();
+start({
+  setupTestIsolationValidation: true,
```
rwjblue: 💯 👍

tests/unit/query-cache-test.js
```diff
@@ -33,10 +33,11 @@ module('unit/query-cache', function(hooks) {
 
     this.queryCache = this.store._queryCache;
     this.adapterAjax = this.sinon.stub(this.adapter, 'ajax').returns(resolve());
-  }),
```
rwjblue: 😱

PR #382, opened 2019-09-04
chore: update travis.yml

hjdivad:
- Fail the build on beta channel regressions
- Update README to clarify support policy now that we are less dependent
  on private Ember Data API


hjdivad:
@igorT LMK when the ember data issues have been backported to ember

PR #384, opened 2019-09-05
Add support for adapterOptions to queryURL

loganrosen:
I've added support for `adapterOptions`, custom options that can be passed along to the adapter, to the `queryURL` method. This matches the behavior of other Ember Store methods, like the ones provided by the REST adapter.


hjdivad:
thanks @loganrosen !

PR #385, opened 2019-09-05
feat: interoperability between Ember Data and m3 in debug-adapter

SYU15:
This PR will allow both `DS` models and `m3` models to render and be tracked in `Ember Inspector`, which will help developers as they migrate onto `m3`.

### DIFF
--- a/addon/adapters/interop-debug-adapter.js
+++ b/addon/adapters/interop-debug-adapter.js
@@ -1,0 +1,74 @@
+import { DebugAdapter } from 'ember-data/-private';
+import M3DebugAdapter from './m3-debug-adapter';
+import { get, defineProperty } from '@ember/object';
+import { inject } from '@ember/service';
+import { default as MegamorphicModel } from '../model';
+
+/*
+  Extend Ember Data's `DebugAdapter` to handle both m3 and DS.Model model types
+
+  @class InteropDebugAdapter
+  @extends DebugAdapter
+  @private
+*/
+
+export default class InteropDebugAdapter extends DebugAdapter {
+  init() {
+    super.init(...arguments);
+    const store = get(this, 'store');
+    const schema = get(this, 'schema');
+    this._m3DebugAdapter = M3DebugAdapter.create({
+      store,
+      schema,
+    });
+  }
+
+  /**
+    Function to call m3's debug adapter addedType method
+    This is only needed for m3 records that have been dynamically added after
+    watchModelTypes has been called
+    @private
+    @method addedType
+    @param {String} type Model type
+  */
+  addedType(type) {
+    this._m3DebugAdapter.addedType(type);
+  }
+
+  /**
+    Calls the getRecordColumnValues function from either m3's debug adapter
+    or Ember Data's debug adapter depending on the record type
+    @public
+    @method getRecordColumnValues
+    @param {MegamorphicModel|DS.Model} record to get values from
+    @return {Object} Keys should match column names defined by the model type
+  */
+  getRecordColumnValues(record) {
+    if (record instanceof MegamorphicModel) {
+      return this._m3DebugAdapter.getRecordColumnValues(record);
+    }
+    return super.getRecordColumnValues(record);
+  }
+
+  /**
+    Calls both the m3 and Ember Data watchModelTypes functions
+    @public
+    @method watchModelTypes
+    @param {Function} typesAdded Callback to call to add types
+    Takes an array of objects containing wrapped types
+    @param {Function} typesUpdated Callback to call when a type has changed
+    Takes an array of objects containing wrapped types
+    @return {Function} Method to call to remove all observers from m3 and DS.Model model types
+  */
+  watchModelTypes(typesAdded, typesUpdated) {
+    let releaseM3 = this._m3DebugAdapter.watchModelTypes(typesAdded, typesUpdated);
+    let releaseSuper = super.watchModelTypes(typesAdded, typesUpdated);
+
+    return () => {
+      releaseSuper();
+      releaseM3();
+    };
+  }
+}
+
+defineProperty(InteropDebugAdapter.prototype, 'schema', inject('m3-schema'));
--- a/tests/integration/interop-debug-adapter-test.js
+++ b/tests/integration/interop-debug-adapter-test.js
@@ -1,0 +1,356 @@
+import { module, test } from 'qunit';
+import sinon from 'sinon';
+import { setupTest } from 'ember-qunit';
+import DS from 'ember-data';
+import { settled } from '@ember/test-helpers';
+import { capitalize, underscore } from '@ember/string';
+import { A } from '@ember/array';
+import InteropDebugAdapter from 'ember-m3/adapters/interop-debug-adapter';
+import DefaultSchema from 'ember-m3/services/m3-schema';
+
+const BOOK_MODEL_TYPE = 'com.example.bookstore.Book';
+const NEW_MODEL_TYPE = 'com.example.newModel';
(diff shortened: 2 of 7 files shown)


addon/adapters/debug-adapter.js
```diff
@@ -1,8 +1,7 @@
-import { A, isArray } from '@ember/array';
-import DataAdapter from '@ember/debug/data-adapter';
+import { DebugAdapter as DataAdapter } from 'ember-data/-private';
```
igorT: Lets maybe not do the rename as the Ember.js provided class is called DataAdapter, so this might be confusing.

addon/adapters/interop-debug-adapter.js
```diff
@@ -0,0 +1,70 @@
...
+    @public
+    @method watchModelTypes
+    @param {Function} typesAdded Callback to call to add types
+    Takes an array of objects containing wrapped types (returned from `processAddedTypes`)
+    @param {Function} typesUpdated Callback to call when a type has changed
+    Takes an array of objects containing wrapped types
+    @return {Function} Method to call to remove all observers
+  */
+  watchModelTypes(typesAdded, typesUpdated) {
+    this._m3DebugAdapter.watchModelTypes(typesAdded, typesUpdated);
```
SYU15: I'm uncertain as to how I should handle combining the outputs of the release methods, since each of the `watchModelTypes` returns a callback but I think there should be a shared array of release functions to iterate over for both:
Screen Shot 2019-09-09 at 9 47 36 AM

It's unclear to me if I can accomplish this without tweaking how `watchModelTypes` works in [Ember's Data Adapter](https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/extension-support/lib/data_adapter.js#L133). I can, of course, change the logic for the `m3's` adapter as well but need some guidance as to what the path forward should be.

tests/dummy/app/models/cake.js
```diff
@@ -0,0 +1,7 @@
+import DS from 'ember-data';
+
+export default DS.Model.extend({
```
SYU15: Will remove this and the logic in `index.js` once I'm done manually testing.

tests/integration/interop-debug-adapter-test.js
```diff
@@ -0,0 +1,222 @@
...
+const generateColumns = columnArray =>
+  columnArray.map(attribute => ({ name: attribute, desc: attribute }));
+
+module('integration/interop-debug-adapter', function(hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function() {
+    this.sinon = sinon.createSandbox();
+
+    this.Publisher = DS.Model.extend({
```
SYU15: For some reason, `DS` models aren't being recognized in the `debug-adapter` for this test, even though it works when I test it manually. Am I registering the model incorrectly?

tests/integration/m3-debug-adapter-test.js
```diff
@@ -241,51 +244,28 @@ module('integration/debug-adapter', function(hooks) {
...
-    });
-    await settled();
-
-    assert.ok(
-      this.debugAdapter.addedType.calledWithExactly('com.example.new-model'),
-      'addedType is called when new model types are pushed into the store'
-    );
-  });
-
   test('addedType wraps model and notifies Ember Inspector of new types', async function(assert) {
```
SYU15: I think I can remove this test because the test for `watchModelTypes` in `interop-debug-adapter-test.js` handles this case in a better way but let me know if you have any concerns.

tests/dummy/app/models/cake.js
hjdivad: sounds good 👍

tests/integration/m3-debug-adapter-test.js
hjdivad: yep i agree 👍

addon/adapters/interop-debug-adapter.js
hjdivad: Return a function that invokes both callbacks

hjdivad:
```js
watchModelTypes(typesAdded, typesUpdated) {
  let releaseM3 = this._m3DebugAdapter.watchModelTypes(typesAdded, typesUpdated);
  let releaseSuper = super.watchModelTypes(typesAdded, typesUpdated);

  return () => {
    releaseSuper();
    releaseM3();
  }
}
```

tests/integration/interop-debug-adapter-test.js
hjdivad: @SYU15 I'm not sure what you mean; CI is 🍏 so I presume your tests pass; can you add a failing test that illustrates the way in which `Publisher` is not being recognized?

addon/adapters/interop-debug-adapter.js
SYU15: Ohhh, ok seems obvious now that you said it 🤦‍♀Will do that, thanks!

tests/integration/interop-debug-adapter-test.js
SYU15: Sure will update the test to fail when I'm back from DTO

addon/adapters/interop-debug-adapter.js
hjdivad: np! 😀

I wouldn't worry about catching errors or anything unless we really need to ensure `releaseM3` runs or something; i would expect a simple composition here should be fine and let any errors get thrown.

tests/integration/interop-debug-adapter-test.js
hjdivad: thanks @SYU15!

SYU15: I've updated the test to have the expected value for `typesAdded`, which is failing. Lmk what you think.

addon/adapters/interop-debug-adapter.js
```diff
@@ -0,0 +1,75 @@
...
+    let releaseSuper = super.watchModelTypes(typesAdded, typesUpdated);
+
+    return () => {
+      releaseSuper();
+      releaseM3();
+    };
+  }
+}
+
+defineProperty(DebugAdapter.prototype, 'schema', inject('m3-schema'));
```
igorT: `InteropDebugAdapter`

tests/dummy/app/routes/index.js
```diff
@@ -27,6 +27,18 @@ export default Route.extend({
           },
         },
       });
+
+      this.store.push({
```
igorT: remove this as well before merging

tests/integration/interop-debug-adapter-test.js
```diff
@@ -0,0 +1,350 @@
...
+
+    await settled();
+    assert.ok(
+      this.interopDebugAdapter.addedType.calledWithExactly('com.example.new-model'),
+      'addedType is called when new model types are pushed into the store'
+    );
+  });
+
+  test('watchModelTypes correctly watches both m3 and DS.Model records', async function(assert) {
+    let typesAddedCallCount = 0;
```
igorT: would be great to great `assert.expects(5)`

tests/integration/interop-debug-adapter-test.js
```diff
@@ -0,0 +1,350 @@
...
+        attributes: {
+          name: 'Horror',
+          description: 'Fiction that will keep you up at night.',
+        },
+      },
+    });
+  });
+
+  test('typesUpdated is called when new m3 and DS.Model records are added of an existing type', async function(assert) {
+    let typesUpdatedCallCount = 0;
```
igorT: also here

PR #426, opened 2019-10-11
Add "meta" support for queryURL

2hu12:


addon/query-cache.js
```diff
@@ -239,14 +240,15 @@ export default class QueryCache {
...
-  _createQueryArray(internalModelsOrModels, query) {
+  _createQueryArray(internalModelsOrModels, query, payload) {
     let array = M3QueryArray.create({
       modelName: '-ember-m3',
       store: this._store,
       manager: this._recordArrayManager,
 
       queryCache: this,
       query,
+      meta: assign({}, payload.meta),
```
igorT: Let's pass in meta directly, and not assign. Is there are reason we would want the base case to be empty object?

igorT:
Thanks! Other than the comment, looks good to me

addon/query-cache.js
2hu12: Just borrow this from https://github.com/emberjs/data/blob/681d922fe6736df0acab33f3eecfceba91cb8e65/packages/store/addon/-private/system/record-array-manager.js#L243, I guess it would make sure that the `meta` is always an object, or `payload.meta` could be passed in derectly if it doesn't matter.

igorT: ah makes sense :+1:

PR #428, opened 2019-10-15
Revert DataAdapter interop

runspired:
This reverts #385 

Prior to landing the original PR work was done in EmberData to eliminate the entanglement of the DataAdapter in EmberData with `@ember-data/model`. That work has not been released as a canary yet, so CI for #385  passed when it should have failed.

The changes in EmberData introduce two complicating factors.

First, the private import path changes. This turns out to be the easier bit to handle, and can be done using the following pattern:

```js
import require, { has } from 'require';

let DebugAdapter;
if (has('@ember-data/debug')) {
  DebugAdapter = require('@ember-data/debug').default;
} else {
  DebugAdapter = require('ember-data/-private').DebugAdapter;
}
```

Second, the implementation is now agnostic to what model is use. It was authored with the intent of working for *any* Model implementation. This turns out to be problematic for the approach in the original PR here as there was an implicit assumption that ember-data would filter to only models extending `@ember-data/model` and also that ember-data would only watch the types of models that were statically available (via a scan of `requirejs.entries`). Neither of these assumptions is true any longer.


runspired:
Identified a path to an easy fix with @igorT and closing in favor of #429

PR #429, opened 2019-10-15
[BUGFIX] interop with newer ember-data

runspired:
Replaces #428

### DIFF
--- a/addon/adapters/interop-debug-adapter.js
+++ b/addon/adapters/interop-debug-adapter.js
@@ -1,8 +1,15 @@
-import { DebugAdapter } from 'ember-data/-private';
 import M3DebugAdapter from './m3-debug-adapter';
 import { get, defineProperty } from '@ember/object';
 import { inject } from '@ember/service';
 import { default as MegamorphicModel } from '../model';
+import require, { has } from 'require';
+
+let DebugAdapter;
+if (has('@ember-data/debug')) {
+  DebugAdapter = require('@ember-data/debug').default;
+} else {
+  DebugAdapter = require('ember-data/-private').DebugAdapter;
+}
 
 /*
   Extend Ember Data's `DebugAdapter` to handle both m3 and DS.Model model types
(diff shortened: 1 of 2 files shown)


tests/integration/interop-debug-adapter-test.js
```diff
@@ -226,10 +232,21 @@ module('integration/interop-debug-adapter', function(hooks) {
...
             'Correct type object passed into typesAdded for DS.Model record types'
           );
         case 3:
+          if (HasDebugAdapterPackage) {
+            return assert.deepEqual(
+              typesToSend,
+              dsTypesAdded2,
+              'Correct type object passed into typesAdded for DS.Model record types'
+            );
+          } else {
```
igorT: don't need an empty else

tests/integration/interop-debug-adapter-test.js
```diff
@@ -147,7 +150,7 @@ module('integration/interop-debug-adapter', function(hooks) {
   });
 
   test('watchModelTypes correctly watches both m3 and DS.Model records', async function(assert) {
-    assert.expect(6);
+    assert.expect(HasDebugAdapterPackage ? 7 : 6);
```
igorT: :+1:

tests/integration/interop-debug-adapter-test.js
```diff
@@ -7,6 +7,9 @@ import { capitalize, underscore } from '@ember/string';
 import { A } from '@ember/array';
 import InteropDebugAdapter from 'ember-m3/adapters/interop-debug-adapter';
 import DefaultSchema from 'ember-m3/services/m3-schema';
+import { has } from 'require';
+
+const HasDebugAdapterPackage = has('@ember-data/debug');
```
igorT: published the canary so we can do a version check now

addon/adapters/interop-debug-adapter.js
```diff
@@ -71,4 +82,13 @@ export default class InteropDebugAdapter extends DebugAdapter {
   }
 }
 
+function interceptDataTypes(schema, method) {
```
igorT: a comment here as to why we need to do this would be great

tests/integration/interop-debug-adapter-test.js
igorT: a comment here describing why case 3 happens would also be 👍

igorT:
Left a couple comments, but gonna merge this as is so we can unblock ED

PR #451, opened 2019-11-12
Add headers support for queryURL

pyuan:
According to the m3 docs, it looks like currently in order to set custom headers, we have to reference the -ember-m3 adapter directly, it would be nice if we are able to pass the headers through store.queryURL so we can still take advantage of the serializer in queryURL: https://github.com/hjdivad/ember-m3#serializer--adapter

This change allows custom headers to be passed into queryURL. @runspired @hjdivad

### DIFF
--- a/addon/query-cache.js
+++ b/addon/query-cache.js
@@ -126,6 +126,9 @@ export default class QueryCache {
     if (options.params) {
       ajaxOptions.data = options.params;
     }
+    if (options.adapterOptions && options.adapterOptions.headers) {
+      ajaxOptions.headers = options.adapterOptions.headers;
+    }
     return adapter.ajax(url, method, ajaxOptions);
   }
 
--- a/tests/unit/query-cache-test.js
+++ b/tests/unit/query-cache-test.js
@@ -192,6 +192,29 @@ module('unit/query-cache', function(hooks) {
     });
   });
 
+  test('.queryURL can accept headers', function(assert) {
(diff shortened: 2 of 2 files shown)


tests/unit/query-cache-test.js
```diff
@@ -192,6 +192,29 @@ module('unit/query-cache', function(hooks) {
...
+    return this.queryCache
+      .queryURL('/uwot', { adapterOptions: { headers: { param: 'value' } } })
+      .then(() => {
+        assert.deepEqual(
+          stubCalls(this.adapterAjax),
+          [[this.adapter + '', ['/uwot', 'GET', { headers: { param: 'value' } }]]],
+          'adapter.ajax called with right headers'
+        );
+      });
+  });
```
hjdivad: let's add a testing make sure that these still get merged with any headers on `adapter.headers`

addon/query-cache.js
```diff
@@ -126,6 +126,9 @@ export default class QueryCache {
     if (options.params) {
       ajaxOptions.data = options.params;
     }
+    if (options.adapterOptions && options.adapterOptions.headers) {
```
hjdivad: We need to pass the headers in the other return when the adapter specifies its own `queryURL` (also needs a test for that case)

runspired: we need a test but we are already passing the headers because we are passing all of options

pyuan: added test when adapter specifies its own queryURL

hjdivad:
Thanks @pyuan !

PR #453, opened 2019-11-14
Strip FeatureFlagged ED code

igorT:


package.json
```diff
@@ -58,6 +58,8 @@
     ]
   },
   "dependencies": {
+    "@ember-data/private-build-infra": "^3.15.0-beta.0",
```
rwjblue: I don't think we should have a dep here, can you remove

index.js
```diff
@@ -1,14 +1,44 @@
...
+      features = this.project.require('@ember-data/private-build-infra/src/features');
+    } catch (e) {
+      features = { CUSTOM_MODEL_CLASS: false };
+    }
+
     this.options = this.options || {};
     this.options.babel = this.options.babel || {};
+    // this ensures that the same `@ember-data/canary-features` processing that the various
+    // ember-data addons do is done in the dummy app
+    this.options.babel.plugins = [...debugMacros(features)];
```
hjdivad: seems like we should not stomp existing `options.babel.plugins`

PR #459, opened 2019-11-18
Fix canary feature imports

igorT:
Import from ember-data for features was incorrect which failed once newest ED was published

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,7 +28,7 @@ module.exports = {
 
     let features;
     try {
-      features = this.project.require('@ember-data/private-build-infra/src/features');
+      features = this.project.require('@ember-data/private-build-infra/src/features')();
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -28,7 +28,7 @@ module.exports = {
 
     let features;
     try {
-      features = this.project.require('@ember-data/private-build-infra/src/features');
+      features = this.project.require('@ember-data/private-build-infra/src/features')();
```
rwjblue: Ya, FWIW, this changed in the most recent set of canaries.

I'm concerned that our `try`/`catch` didn't properly absorb this though (and fallback to false)...

igorT: It didn't absorb because it didn't error. Import worked, but made features equal to a function, so there was no catch

PR #472, opened 2019-12-03
fix: remove seen model types in debug-adapter when inspector is closed

SYU15:
This fix is similar to this PR in Ember Data: https://github.com/emberjs/data/pull/6549 which ensures that when the inspector is closed and reopened, the adapter correctly rewatches previously added model types.

### DIFF
--- a/addon/adapters/m3-debug-adapter.js
+++ b/addon/adapters/m3-debug-adapter.js
@@ -278,6 +278,9 @@ export default class M3DebugAdapter extends DataAdapter {
     let release = () => {
       this.localReleaseMethods.forEach(fn => fn());
       this.releaseMethods.removeObject(release);
+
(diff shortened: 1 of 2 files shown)


addon/adapters/m3-debug-adapter.js
```diff
@@ -278,6 +278,9 @@ export default class M3DebugAdapter extends DataAdapter {
     let release = () => {
       this.localReleaseMethods.forEach(fn => fn());
       this.releaseMethods.removeObject(release);
+
+      // Clear out the model types in seenTypesInAdapter so they can be added again if the inspector is re-opened
+      this.seenTypesInAdapter.clear();
```
hjdivad: Can we add a test for this?

hjdivad:
thanks @SYU15 !

PR #477, opened 2019-12-11
perf: instantiate errors lazily

runspired:
Noticed this while implementing some build infra. Currently we greedily create this for every record when it'll only be used for a tiny number. Given the heavy weight of the Errors class I suspect we will see a significant improvement in render times for applications using M3.

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -78,7 +78,7 @@ export default class MegamorphicModel extends EmberObject {
 
     this._topModel = this._topModel || this;
     this._parentModel = this._parentModel || null;
-    this._errors = DS.Errors.create();
+    this._errors = null;
     this._init = true;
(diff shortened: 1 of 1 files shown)


addon/model.js
```diff
@@ -514,6 +514,9 @@ export default class MegamorphicModel extends EmberObject {
   // Errors hash that will get update,
   // upon validation errors
   get errors() {
+    if (this._errors === null) {
```
igorT: I think the reason for this was that errors is sometimes set/manipulated by addons so we shouldn't be messing with that if thats the case. Might be safer to still make _errors lazy and stash in __errors

runspired: Since there is no setter for `errors` it was not previously possible to set it except by setting `_errors` which would still work here.

PR #478, opened 2019-12-12
feat: build infra

runspired:
This PR

- Refactors tests so that all interop tests are in one location
   - removes those tests when the EmberData package is not present
- Adds infra for stripping code based on presence of ember-data packages
- Improves our infra for stripping based on features
  - Removes our infra code from production builds
- adds CI test coverage for scenarios in which not all of EmberData is present
- wraps package specific code for removal if the package isn't there to interop with
- Causes M3 to bring what EmberData packages it needs
- Adds a `-json-api` fallback adapter to the dummy app (to replace the fallback we were hitting from the adapter package) and makes it error for everything but the ajax request it was making to a fixture
- Fixes how we configure the store for M3 to not use reopen if not using all of EmberData.

Since we don't use ember-auto-import this PR removes it. This is to work around https://github.com/ef4/ember-auto-import/issues/189#issuecomment-462985416

Because we have both ember-qunit and qunit in dependencies, if we bring back ember-auto-import we will need to configure the `exclude` escape hatch: https://github.com/ef4/ember-auto-import#customizing-build-behavior


package.json
```diff
@@ -60,10 +60,10 @@
   "dependencies": {
     "babel-plugin-debug-macros": "^0.3.3",
     "ember-cli-babel": "^7.13.0",
-    "ember-compatibility-helpers": "^1.2.0-beta.1"
-  },
-  "peerDependencies": {
-    "ember-data": ">= 3.5.1"
+    "@ember-data/store": "^3.16.0-alpha.2",
```
stefanpenner: If users have different versions of ember-data, strange duplication will occur. A solution to that is a pre-req to this approach.

addon/adapters/interop-debug-adapter.js
```diff
@@ -2,18 +2,22 @@ import M3DebugAdapter from './m3-debug-adapter';
...
 import { default as MegamorphicModel } from '../model';
-import require, { has } from 'require';
+import require from 'require';
 import { run } from '@ember/runloop';
+import DataAdapter from '@ember/debug/data-adapter';
+import { HAS_DEBUG_PACKAGE, HAS_EMBER_DATA_PACKAGE } from 'ember-m3/-infra/packages';
 
 let DebugAdapter;
-if (has('@ember-data/debug')) {
+if (HAS_DEBUG_PACKAGE) {
```
stefanpenner: It would be interesting to see if @eaf4's `@embroider/macro` library could satisfy these needs, if not we should reach out to him to ensure it does.

But I suspect it can today, via a combination of `moduleExists` and `importSync`. See: https://github.com/emberjs/rfcs/pull/507/files#diff-dee406d92e20c47ccea1291cc51ba8f8R550

Using the `@embroider/macros` is not a blocker, but knowing it can or cannot be used should be considered a blocker. If it cannot be used, let's articulate why, and work with @eaf4 to ensure in the future how ember-m3 / ember-data and embroider want to work, are compatible.

IMHO nothing here seems fundamentally incompatible, but I want to see if we can start either being compatible immediately, or informing parallel efforts to ensure we will soon be.

stefanpenner: For reference, the macro system can be used outside of embroider in classic apps as well via: https://www.npmjs.com/package/@embroider/macros

src/debug-macros.js
```diff
@@ -0,0 +1,150 @@
+/* eslint-env node */
+'use strict';
+const requireEsm = require('esm')(module);
```
stefanpenner: ESM typically slows stuff down quite abit, is it required?

src/debug-macros.js
```diff
@@ -0,0 +1,150 @@
...
+const semver = require('semver');
+
+function gte(availableVersion, compatVersion) {
+  return semver.gte(semver.minVersion(availableVersion), semver.minVersion(compatVersion));
+}
+
+function getPackage(dep, packageName, seen) {
```
stefanpenner: I didn't see tests for the functions in this module, can you point me to them? If they don't exist, they should be implemented.

tests/dummy/app/adapters/-json-api.js
```diff
@@ -0,0 +1,47 @@
+import EmberObject from '@ember/object';
+
+export default class Adapter extends EmberObject {
```
stefanpenner: tests?

tests/interop/global-cache-test.js
```diff
@@ -0,0 +1,194 @@
...
+import { setupTest } from 'ember-qunit';
+import { run } from '@ember/runloop';
+import DefaultSchema from 'ember-m3/services/m3-schema';
+import sinon from 'sinon';
+import { CUSTOM_MODEL_CLASS } from 'ember-m3/-infra/features';
+import { HAS_MODEL_PACKAGE } from 'ember-m3/-infra/packages';
+import require from 'require';
+
+let Model, attr;
+if (HAS_MODEL_PACKAGE) {
```
stefanpenner: Since this test can run with different `Model` and `attr` implementations, do the tests run or need to run multiple combinations/permutations?

package.json
runspired: This PR contains a solution to this already. If the app already contains EmberData, we exclude our own versions of the addons.

addon/adapters/interop-debug-adapter.js
runspired: it's not clear to me from reading those docs that `importSync` and some of the other macros are compatible outside of embroider given the extensive knowledge of the build they need. As the embroider ecosystem matures I'm happy to port to newer shared solutions, but given debug-macros is the standard in the community I'd prefer to stick to well trodden paths.

cc @rwjblue for thoughts.

src/debug-macros.js
runspired: It's required if we want to have unstripped flags at runtime or ever want typescript support, but since I configured them to always strip though I'm thinking of eliminating the `-infra` directory entirely as debug-macros uses a virtualized module.

tests/dummy/app/adapters/-json-api.js
runspired: Tests for tests don't make sense?

tests/interop/global-cache-test.js
runspired: They don't. `Model` and `attr` will always be the same, only their source module will change.

src/debug-macros.js
runspired: virtualized them

addon/adapters/interop-debug-adapter.js
stefanpenner: My recommendation is not to switch, my recommendation is to confirm this is going down that path that could be plausibly implemented with `@embroider/macros`. We want to avoid, going down a path that isn't aligned ecosystem wide.

My reading of the macros and this code suggests that the approaches are similar enough, that taking this approach today doesn't put us on a fundamentally incompatible path.

stefanpenner: But if this approach is being take, please be aware near term work will be requested to ensure this is compatible with embroider. So please prepare for that.

package.json
stefanpenner: Can you share a link to the code that implements that?

runspired: https://github.com/hjdivad/ember-m3/pull/478/files#diff-168726dbe96b3ce427e7fedce31bb0bcR24-R37

runspired: The detection is a bit more raw than our other detection logic in the other hooks because ember-cli calls this hook prior to setting up addons, if you can think of ways to improve it I'm all 👂

stefanpenner: that link doesn't appear to work for me..

index.js
```diff
@@ -1,43 +1,62 @@
...
-  ];
-
-  return plugins;
+/*
+ shouldIncludeChildAddon runs during
+ addon init so does not have access to
+ the host application instance.
+*/
+function hasDataPackage(addon) {
+  let addons = addon.project.addonPackages;
```
stefanpenner: If possible using `project.findAddonByName` may be a more resilient approach as technically ember-cli has some addon initialization that occurs: https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js#L634

(I added this, specifically for this use-case)

index.js
```diff
@@ -1,43 +1,62 @@
...
+  return addons['ember-data'] !== undefined;
 }
 
 module.exports = {
   name: 'ember-m3',
```
stefanpenner: Although it hasn't been done so far, it may be worth adding a guard erroring if more then 1 ember-m3 is installed. Enforcing the highlander stuff. Otherwise, if two are installed, and they each bring their own ember-data stuff with, sadness will occur.

index.js
```diff
@@ -1,43 +1,62 @@
...
+    let app = this._findHost();
 
     this.options = this.options || {};
     this.options.babel = this.options.babel || {};
-    // this ensures that the same `@ember-data/canary-features` processing that the various
-    // ember-data addons do is done in the dummy app
-    this.options.babel.plugins = [...debugMacros(features)];
+    let plugins = this.options.babel.plugins;
+    let newPlugins = getDebugMacros(app);
+    this.options.babel.plugins = Array.isArray(plugins) ? plugins.concat(newPlugins) : newPlugins;
```
stefanpenner: Inserting these debug macros globally  is surprisingly, do we not want to only run them for the subset that need them?

index.js
```diff
@@ -1,43 +1,62 @@
...
 
-    let features;
-    try {
-      features = this.project.require('@ember-data/private-build-infra/src/features')();
-    } catch (e) {
-      features = { CUSTOM_MODEL_CLASS: false };
+    if (isProd) {
+      exclude.push('adapters');
     }
+    tree = new Funnel(tree, {
```
stefanpenner: This creates an un-needed funnel if we are in the else clause of the `isProd` guard.
A potential alternative would be:

```js
if (isProd) {
  tree = new Funnel(tree, { exclude, })
}
```

addon/query-cache.js
```diff
@@ -5,7 +5,7 @@ import { assign } from '@ember/polyfills';
 
 import MegamorphicModel from './model';
 import M3QueryArray from './query-array';
-import { CUSTOM_MODEL_CLASS } from './feature-flags';
+import { CUSTOM_MODEL_CLASS } from 'ember-m3/-infra/features';
```
hjdivad: p3/3 this import should actually resolve to a real path, as it does in master and on ember data.

Right now `-infra/features` seems to be purely synthetic.

runspired: I removed the real path as it was unnecessary

addon/model.js
```diff
@@ -171,7 +255,7 @@ export default class MegamorphicModel extends EmberObject {
       // TODO need to walk the chain down as well to notify changes
     }
     if (this !== this._topModel) {
-      this._topModel._updateCurrentState(state);
+      this._topModel._updateCurrentState(!CUSTOM_MODEL_CLASS && state);
```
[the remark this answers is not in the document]
runspired: Never, it’ll be stripped entirely if false. Also in the false branch this method takes no args

addon/model.js
```diff
@@ -37,6 +34,93 @@ if (!CUSTOM_MODEL_CLASS) {
...
+  deletedUncommitted = dUncommitted;
+  loadedSaved = lSaved;
+  updatedUncommitted = upUncommitted;
+}
+
+function isInvalidError(error) {
+  if (HAS_ADAPTER_PACKAGE) {
+    return error && error instanceof InvalidError;
+  } else {
+    return error && error.isAdapterError === true && error.code === 'InvalidError';
```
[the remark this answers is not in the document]
runspired: EmberData does.

addon/model.js
```diff
@@ -37,6 +34,93 @@ if (!CUSTOM_MODEL_CLASS) {
...
+      }
+      if (DEBUG && !RootState) {
+        throw new Error(`ember-m3 requires the @ember-data/model package to import RootState`);
+      }
+    } else {
+      // it is possible RootState still lives in the store package
+      RootState = require('@ember-data/store/-private').RootState;
+      if (!RootState) {
+        throw new Error(`ember-m3 requires the @ember-data/model package to import RootState`);
+      }
```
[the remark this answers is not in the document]
runspired: Errors is private and as such has been subject to a few locations during refactoring. This makes sure we check the most likely to the least likely before erring if not available

CHANGELOG.md
```diff
@@ -182,7 +182,7 @@
 
 ## 0.7.5
 
-- Model supports client validation errors. `errors` an instance of DS.Errors provides list of vaidation errors.
+- Model supports client validation errors. `errors` an instance of Errors provides list of vaidation errors.
```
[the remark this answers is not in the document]
runspired: `DS.` doesn't have meaning in modern EmberData, and `DS.Errors` was a private class. Near term we should refactor away from it, short term I'll make the docs a bit more clear it is the same "errors" object as @ember-data/model errors

addon/model.js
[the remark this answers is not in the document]
hjdivad: @runspired where?  I see it only in the [adapter package](https://github.com/emberjs/data/blob/master/packages/adapter/addon/error.js#L75) but this branch is specifically when we do not have this package.

addon/query-cache.js
hjdivad: @runspired the real path should exist, as it does in ember data and ember.  Without it, the code ends up being much harder to follow when it has "phantom" imports.

hjdivad:
> tests are very red with custom model class enabled

ok per OOB ember data doesn't support custom model being enabled in isolation; this is expected and the action item here is for ED to hard fail with a more informative message

index.js
runspired: @stefanpenner what do you mean by global? this.options.babel will apply only to the addon, not the consuming app.

addon/adapters/interop-debug-adapter.js
```diff
@@ -2,18 +2,22 @@ import M3DebugAdapter from './m3-debug-adapter';
...
   DebugAdapter = require('@ember-data/debug').default;
-} else {
+} else if (HAS_EMBER_DATA_PACKAGE) {
   DebugAdapter = require('ember-data/-private').DebugAdapter;
+} else {
+  DebugAdapter = DataAdapter;
```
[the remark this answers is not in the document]
runspired: My intent is to follow up with stripping this in the "no full compat" case and registering our primary data adapter instead.

addon/model.js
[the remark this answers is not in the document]
runspired: @hjdivad the check in the `else` is the same check we do in EmberData now. We made this change to support folks not needing to use `InvalidError` the class but as an interface instead.

I maintained the `InvalidError` check here for the case where we have the ability to make the instanceof check for a wider compat range.

src/debug-macros.js
runspired: Added tests and made them run in travis

index.js
runspired: I investigated this and that util won't work for us because it does not recurse the addon dep tree, otherwise it would be very helpful.

index.js
runspired: I'll add that in a follow up 👍

addon/adapters/interop-debug-adapter.js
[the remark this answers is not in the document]
hjdivad: why would we not `HAS_DEBUG_PACKAGE` when `@ember-data/debug` is in `dependencies`?

addon/model.js
```diff
@@ -37,6 +34,93 @@ if (!CUSTOM_MODEL_CLASS) {
...
+  /*
+    Get access to the Private Errors class
+  */
+  if (HAS_MODEL_PACKAGE) {
+    // Errors is supposed to live in the model package
+    Errors = require('@ember-data/model/-private').Errors;
+    if (!Errors) {
+      // it is possible Errors still lives in the store package
+      Errors = require('@ember-data/store/-private').Errors;
+    }
```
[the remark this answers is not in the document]
hjdivad: This doesn't look resolved to me; it's still not clear what specific versions of Ember Data these branches are for.

addon/adapters/interop-debug-adapter.js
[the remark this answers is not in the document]
runspired: We exclude our version when we detect EmberData already present, and we may have an EmberData version that doesn't have that package.

addon/model.js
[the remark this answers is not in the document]
runspired: I'll add even more comments

package.json
```diff
@@ -58,13 +60,16 @@
     ]
   },
   "dependencies": {
+    "@ember-data/debug": "^3.16.0-alpha.3",
+    "@ember-data/model": "^3.16.0-alpha.3",
+    "@ember-data/store": "^3.16.0-alpha.3",
```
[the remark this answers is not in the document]
runspired: @hjdivad it's the `shouldIncludeChildAddon` hook in `index.js`. I appear to not be able to give folks working links, I think because I'm part of a github beta trial for a different kind of link.

addon/model.js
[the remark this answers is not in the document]
hjdivad: If I run
```
EMBER_DATA_FEATURE_OVERRIDE=CUSTOM_MODEL_CLASS,REQUEST_SERVICE,IDENTIFIRS,RECORD_DATA_STATE,RECORD_
DATA_ERRORS node $(volta which ember) s --no-live-reload --port=0

I see `_updateCurrentState` called with `false` for test `3867eb4c` (`unit/model/state: isDirty`).

Also when we fix this so that we can can toggle these tests in dev mode the code won't be stripped at all, which is required for the runtime toggling to work.

```

index.js
```diff
@@ -1,62 +1,61 @@
...
+  shouldIncludeChildAddon(addon) {
+    if (!hasDataPackage(this)) {
+      return true;
+    }
+    if (addon.name.startsWith('@ember-data')) {
+      /*
+      console.log(
+        `⚠️  ember-m3 is excluding ${addon.name} version ${addon.pkg.version} from the build`
+      );
+      */
```
hjdivad: dead code

index.js
```diff
@@ -1,62 +1,61 @@
...
   included() {
-    this._super.included.apply(this, arguments);
+    this._super.included.call(this, ...arguments);
+    this.configureBabelOptions();
+  },
+
+  shouldIncludeChildAddon(addon) {
+    if (!hasDataPackage(this)) {
+      return true;
+    }
```
hjdivad: Isn't it more complicated than this?

If a user has their own `@ember-data/store` shouldn't that trump the  version we implicitly bring along?

addon/model.js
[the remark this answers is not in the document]
runspired: Added

addon/model.js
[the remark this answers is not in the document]
runspired: Calling it with false is safe (and also the current behavior elsewhere, I simply fixed a few places it had been forgotten)

index.js
runspired: @hjdivad we could make it more complicated in the future if we want. For now I don't think I would recommend it. Detecting is difficult enough as is.

index.js
runspired: @hjdivad should I delete it or make it always print? It can be useful for seeing what's going on.

stefanpenner:
Although this is functional, the current approach may have issues with embroider, maybe if embroider disables all optimizations, but that is counter productive. As implemented this will cause some partners working on that effort extra grief.

---

As implement landing this as-is may be __ok__ as a very short term unblocking step, but we should be good citizens asap. This citizenship would be to attempt to use `@embroider/macros`, and either switch to it or provide actionable feedback to that effort describing what is preventing adoption. cc @eaf4

---

I believe `@embroider/macros` (at master, but we can get @eaf4 to cut a release for us) should be sufficiently to support most use-cases (maybe not all) and should work both with embroider and classic builds via:

* importSync: https://github.com/emberjs/rfcs/blob/embroider/text/0000-embroider-v2-package-format.md#javascript-macro-importsync
* macroCondition  https://github.com/emberjs/rfcs/blob/embroider/text/0000-embroider-v2-package-format.md#javascript-macro-macrocondition
* moduleExists https://github.com/emberjs/rfcs/blob/embroider/text/0000-embroider-v2-package-format.md#javascript-macro-moduleexists
* dependencySatisfies https://github.com/emberjs/rfcs/blob/embroider/text/0000-embroider-v2-package-format.md#javascript-macro-dependencysatisfies
* failBuild: https://github.com/emberjs/rfcs/blob/embroider/text/0000-embroider-v2-package-format.md#javascript-macro-failbuild

```js
import { moduleExists, macroCondition, importSync } from "@ember/macros";
if (macroCondition(moduleExists("ember-data"))) {
  const DS = importSync("ember-data").default;
  DS.Adapter.extend({
    //
  });
}
```

addon/model.js
```diff
@@ -37,6 +34,110 @@ if (!CUSTOM_MODEL_CLASS) {
...
+      // but in some versions is not yet in it's home.
+      // What version this becomes true in is TBD, likely to be 3.17
+      RootState = require('@ember-data/model/-private').RootState;
+      if (!RootState) {
+        // it is possible RootState still lives in the store package
+        // Before migrating to the model package RootState was within the store
+        // This was versions 3.12 -> ~3.16 (TBD)
+        RootState = require('@ember-data/store/-private').RootState;
+      }
+      if (DEBUG && !RootState) {
```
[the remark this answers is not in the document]
runspired: Both should

addon/model.js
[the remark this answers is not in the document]
runspired: The obvious thing is that when the flag is false there is no currentState at all since we don't rely on InternalModel. We won't use the arg, and won't set anything.

addon/model.js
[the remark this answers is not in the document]
runspired: Fixed

index.js
runspired: Added this detection

PR #479, opened 2019-12-12
chore: refactor store mixin

runspired:
The current structure in which we reopen the store is a smell, but was done to present a "no-config" story for the case where the store was extended by a consuming application.

This keeps the reopen when `ember-data` is in use but prepares us for our own `store` service when `ember-data` is not in use.


igorT:
@runspired over discord you mentioned something about import cycles? Is this PR just a refactor or does it also fix issues?

runspired:
@igorT we have a cycle we can't delete as long as we are using reopen. This PR preserves it but makes it slightly more obvious by making it dependent on clearly side-effect imports in two locations:

initializers/m3-store.js imports the store mixin file without using it
test-helper.js imports the initializer without using it

Once we no longer support the usage pattern requiring store reopen we can drop both import statements.

addon/services/store.js
```diff
@@ -0,0 +1,18 @@
...
+
+const HAS_EMBER_DATA_PACKAGE = has('ember-data');
+
+if (!HAS_EMBER_DATA_PACKAGE) {
+  Store = require('@ember-data/store').default;
+  ExportedStore = Store.extend(StoreMixin);
+} else {
+  // the re-open in the initializer will handle configuration
+  // in this case
+  ExportedStore = require('ember-data/store').default;
```
hjdivad: We're not running tests on this branch are we?  We should

hjdivad: This won't work until we have our own store service.  I don't mind landing a WIP PR here but only if the unsupported case fails fast (ie a user build should fail if they lack the ember data package until we actually have a store service that supports that case w/o the mixin)

runspired: We are in the infra PR.

PR #483, opened 2019-12-13
Fix CUSTOM_MODEL_CLASS fallback

hjdivad:
We fell back to a value of `false` which would always strip the code.
This commit changes the fallback to `null` when developing the addon,
which leaves the feature flag a runtime concern.

Also move init code to `included` as `isDevelopingAddon()` is not
reliable in `init`.

Co-authored-by: Robert Jackson <me@rwjblue.com>

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,14 +23,14 @@ function debugMacros(features) {
 module.exports = {
   name: 'ember-m3',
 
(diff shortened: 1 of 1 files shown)


hjdivad:
cc @igorT

PR #485, opened 2019-12-13
Fix ember data feature import

hjdivad:
The feature import was based on Ember Data 3.14, which is not the
current canary.

This commit still depends on an unstable import path
`@ember-data/private-build-infra/src/features`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,8 @@
 /* eslint-env node */
 'use strict';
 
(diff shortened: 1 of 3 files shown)


hjdivad:
cc @igorT

hjdivad:
@igorT @rwjblue we should add a public API to `@ember-data/canary-features` so that we can remove the import from a private path.

PR #489, opened 2019-12-17
update ember-cli to 3.14

hjdivad:
- ~~drop jQuery use in tests~~ (keep it for now for tests in 3.8)
- now we target chrome for tests; class code is readable again


hjdivad:
We can ignore the babel-plugin-debug-macros warnings, they're spurious from ember-resolver and fixed in https://github.com/ember-cli/ember-resolver/pull/475

hjdivad:
warnings are also more actionable now https://github.com/ember-cli/babel-plugin-debug-macros/pull/81

PR #513, opened 2020-01-11
Fix isDirty check for parents of nested models

igorT:
https://github.com/hjdivad/ember-m3/pull/377 introduced `hasDirtyAttr()` in order to have better support for state updating when a record has an attribute set.

https://github.com/hjdivad/ember-m3/pull/433 used `hasDirtyAttr()` to check whether to revert the record to a clean state.

However, `hasDirtyAttr` did not account for nestedModels, thus reverting the parentModel to a clean state even when nestedModels had changes.

This commit removves `hasDirtyAttr` and replaces it with existing method `hasChangedAttributes` which already correctly accounts for the childRecordDatas.

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -475,7 +475,7 @@ export default class MegamorphicModel extends EmberObject {
     this._schema.setAttribute(this._modelName, attr, value, schemaInterface);
     schemaInterface._suppressNotifications = priorSuppressNotifications;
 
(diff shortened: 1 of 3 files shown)


igorT:
@rwjblue @hjdivad I assume you wrote `hasDirtyAttr` because you missed `hasChangedAttribtues` and that the difference is not on purpose, because none of the tests fail with this change

PR #530, opened 2020-01-17
Remove dependency on ArrayProxy

igorT:


igorT:
Added tests for Ember.A behavior @runspired asked for

rwjblue:
I also confirmed that ember-m3 wasn't using `arrangedContent` (or the other features of `ArrayProxy`).

Issue #527, opened 2020-01-16
`MegamorphicModel.toString` cannot be called on the prototype


hjdivad:
The ember-inspector expects to be able to call `toString()` on the prototype as part of it walking up the mixin chain to collect properties to display in the right hand pane.

The current implementation assumes that `init` has already been called (in particular that `this._internalModel` is set) and will error otherwise.

The fix is straightfoward: the id getter should return `null` rather than error when there is no `_internalModel`

cc @syu15

PR #544, opened 2020-01-28
Allow .toString to be called on record prototype

SYU15:
The inspector error is no longer thrown when you click on a record.

Screen Shot 2020-01-27 at 8 34 18 PM

I'm not that sure if the test setup is ideal though, so would like feedback on that. I also tried to stub out `init` as a no-op, but then I got a `Maximum call stack size exceeded` error as it bounced between getting the `id` and `unknownProperty.` Is there actually I way I can test before `init` is run?

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -609,6 +609,12 @@ export default class MegamorphicModel extends EmberObject {
   }
 
   toString() {
+    // Check needed for Ember Inspector support:
+    // https://github.com/emberjs/ember-inspector/blob/545e3c1c7a47f7a033025037f6f1e8d1d4c60624/ember_debug/object-inspector.js#L622
+    if (this === this.constructor.prototype) {
+      return 'MegamorphicModel';
+    }
+
     return `<MegamorphicModel:${this.id}>`;
   }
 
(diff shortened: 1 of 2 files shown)


addon/model.js
```diff
@@ -469,6 +469,10 @@ export default class MegamorphicModel extends EmberObject {
   }
 
   get id() {
+    if (!this._init) {
+      return null;
+    }
+
```
hjdivad: We don't want to do this.  It isn't really valid to ask for the `id` of an uninitialized model (or to ask for anything else for that matter).

The reason we're seeing the issue in the inspector is because it calls `toString` on the prototype to get a description of the class name (instead of `constructor.name or similar).

It'll be a lot clearer to handle this in `toString` and not `id`.  Since the result is being used as a sub-header in the properties pane, returning the class name is probably the most useful.

```js
  toString() {
    if(this === this.constructor.prototype) {
      return 'MegamorphicModel';
    }

    return `<MegamorphicModel:${this.id}>`;
  }
```

As far as testing goes i think it's fine for a unit test POV to confirm that `MegamorphicModel.prototype.toString() === 'MegamorphicModel'`

SYU15: Hmm...that makes a lot of sense, also it will render accurately in the Inspector, I believe. I will refactor my approach then. Thanks!

hjdivad:
Thank you @SYU15 !

hjdivad:
fix #527

PR #567, opened 2020-02-21
Drop direct dependencies on Ember Data

hjdivad:
The code meant to ensure conflicting versions were not present didn't
work and the end result is users who already had old versions of
`ember-data` in their dependencies got trolled when they upgraded.

For now drop the dependencies until we have better infrastructure for
highlander and synchronized addons.


hjdivad:
I've left the code in [index.js](https://github.com/hjdivad/ember-m3/blob/43ca0f697f975228f9ef9b0de7ae692cf9e14d23/index.js#L73) in for now but some of the version checking there can probably be dropped in a subsequent pr

igorT:
Lgtm if this worked in an app when tested with older versions of data

PR #570, opened 2020-02-22
Re-export interop-debug-adapter in app as data-adapter

SYU15:
This is to ensure that the m3 `data-adapter` takes precedence over the Ember data's `@ember-data/debug` adapter so the inspector works as intended.

### DIFF
--- a/app/data-adapter.js
+++ b/app/data-adapter.js
@@ -1,0 +1,1 @@
+export { default } from 'ember-m3/adapters/interop-debug-adapter';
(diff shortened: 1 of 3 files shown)


hjdivad:
Looks great; can we also drop the [injection registration](https://github.com/hjdivad/ember-m3/blob/241d5e14837c43f22aaada6ee774a324203a420b/addon/initializers/m3-store.js#L13-L20) since this supersedes it?

hjdivad:
@SYU15

PR #582, opened 2020-03-04
action time

### LINKED ISSUE #291: switch to github actions

hjdivad:
fix #291


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,31 @@
+# https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions
```
rwjblue: I think we should probably also add a job to test "floating dependencies"

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,31 @@
...
+    - cron: '* 14 * * *' # 1400 UTC / 0600 PT
+
+jobs:
+  test-canaries:
+    name: "t:" # rely on matrix for most of name
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        try-scenario: [beta-channel, canary-channel]
```
rwjblue: I think you should probably also run `default` scenario

bin/report-nightly-failure.js
```diff
@@ -0,0 +1,58 @@
+// https://octokit.github.io/rest.js/v17#issues-list
+const { Octokit } = require('@octokit/rest');
+const m = require('moment');
+
+function assert(test, msg) {
+  if (!test) {
+    throw new Error(msg);
+  }
+}
```
rwjblue:
```js
const assert = require('assert');
```

bin/report-nightly-failure.js
```diff
@@ -0,0 +1,58 @@
...
+  let github = new Octokit({
+    auth: token,
+    userAgent: 'hjdivad/ember-m3 nightly tests issue reporter',
+  });
+  global.g = github;
+
+  let issueTitle = getIssueTitle(runId);
+  let issueSearch = await github.search.issuesAndPullRequests({
+    q: `repo:hjdivad/ember-m3 is:issue in:title ${issueTitle}`,
+  });
```
rwjblue: We should tweak this to filter by the user or label here

bin/report-nightly-failure.js
```diff
@@ -0,0 +1,58 @@
...
+  if (issueSearch.data.total_count > 0) {
+    let issueNumber = issueSearch.data.items[0].number;
+    console.log(`Issue ${issueNumber} already exists for run ${runId}`);
+    return;
+  }
+
+  createIssue({ github, runId });
+}
+
+main({ env: process.env });
```
rwjblue: If you did something like:

```js
module.exports = {
  _main: main,
};

if (require.main === module) {
  main({ env: process.env });
}
```

It would run `main` when invoking it, and export `_main` when requiring it. This would allow us to test it.

hjdivad:
https://github.com/hjdivad/ember-m3/pull/587

PR #602, opened 2020-03-11
Fix nightly run

hjdivad:
cron * 14 means every minute of the 14th hour.  GitHub Actions will
limit this to the minimum of every 5 minutes, although it didn't
actually run 12 times presumably because of the length of the jobs
themselves.

Fix the cron schedule so the job only runs once per night.


hjdivad:
testem intermittently failing to connect is kinda annoying. maybe we can add some verbose logging somewhere to dig into this more.

PR #611, opened 2020-03-19
Fix trackedNotifications for destroying records

igorT:


tests/unit/model/tracked-array-test.js
```diff
@@ -188,7 +189,18 @@ module('unit/model/tracked-array', function(hooks) {
...
+        observerWillChange = removeCount;
+      },
+      didChange: (array, index, removeCount) => {
+        observerDidChange = removeCount;
+      },
+    });
+
     run(() => chapter2.unloadRecord());
+    assert.equal(observerWillChange, 1, 'removal of the record notified to be removed');
+    assert.equal(observerDidChange, 1, 'removal of the record notified that it was removed');
```
hjdivad: follow up from OOB convo; let's add a test to cover the userland regression (ie `[]` recomputation after record unloaded from tracked array)

PR #642, opened 2020-04-03
Fix memory leak in fastboot caused by modifying class prototype

igorT:
When there are multiple requests in parallel in fastboot, using an instance level check for guarding the proto modifications does not work, as each instance of the store will modify the globally shared class prototype, and stash the previously modified one. The stashed methods end up creating a chain leading to a memory leak.

### DIFF
--- a/addon/mixins/store.js
+++ b/addon/mixins/store.js
@@ -21,6 +21,8 @@ export const recordDataToRecordMap = new WeakMap();
 export const recordDataToQueryCache = new WeakMap();
 export const recordToRecordArrayMap = new WeakMap();
 
+let hasModifiedStorePrototype = false;
+
 class SchemaDefinition {
   constructor(store, dsModelSchema) {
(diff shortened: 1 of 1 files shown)


addon/mixins/store.js
```diff
@@ -223,10 +224,11 @@ const StoreMixin = {
           let originalRemove = proto.remove;
           proto.__originalRemove = originalRemove;
           proto.remove = function remove(internalModel) {
-            delete store._globalM3Cache[internalModel.id];
+            delete internalModel.store._globalM3Cache[internalModel.id];
             return originalRemove.apply(this, arguments);
```
stefanpenner: Looks like we are still closing over `originalRemove`, thoughts on using `this.__originalRemove` instead?

stefanpenner: Looks like I forgot to submit this, and it has been addressed. Awsome!

PR #684, opened 2020-04-24
feat: bring own ember-data packages

runspired:
Since the diff here is so large I figure I owe a bit of an explanation on how to review this:

Overwhelmingly the "files added" here are due to having added 27 new addons/apps for testing.

In `dependency-tests/tests` and `dependency-tests/_scenario-addons` you will find organized groups of addons and apps. **The only modifications to these addons/apps from the blueprint for a new addon/app is their package.json.**

For apps, we remove `ember-data` from devDependencies. For addons and apps we then add all or parts of ember-data and ember-m3 to dependencies based on the scenario the test app is meant to cover. These should be relatively simple to skim through.

In `dependency-tests/test-all.js` you have the script to test running the build of all these different add-on/app configurations.

In `dependency-tests/expected.js` you have a fixture generator which generates the expected output of these tests based on the current package.json of ember-m3.

In `src/dependency-checker.js` you have the thing doing the actual dependency management logic we need.

In `index.js` in `shouldIncludeChildAddon` you have the result of those checks and the determination of whether to allow our own version of a package to be included in the build.


runspired:
The LTS 3.12 error is due to this PR, we assert min values for packages and there seems to be an issue where we are seeing the existence of the model package from EmberData. I accounted for this in many different scenarios so will need to investigate to discover what is unique about this one.

package.json
```diff
@@ -15,9 +15,9 @@
     "lint": "yarn lint:js",
     "lint:js": "eslint .",
     "start": "ember server --no-live-reload --port=0",
-    "test": "yarn test:ember && yarn test:node",
+    "test": "yarn test:ember",
     "test:ember": "ember test",
-    "test:node": "qunit node-tests/run.js"
+    "test:dependency-management": "node dependency-tests/test-all.js"
```
igorT, line 20: did node-tests go away?

index.js
```diff
@@ -2,104 +2,42 @@
...
-      /*
-      console.log(
-        `⚠️  ember-m3 is excluding ${addon.name} version ${addon.pkg.version} from the build`
+      debug(
+        `\t⚠️  ember-m3 is excluding its own copy of ${addon.name} version ${addon.pkg.version} from the build`
       );
-      */
       return false;
     }
+    debug(`\t✅  including dependency ${addon.name}`);
```
igorT, line 40: Does this run when you build apps?

index.js
```diff
@@ -2,104 +2,42 @@
...
   },
 
   shouldIncludeChildAddon(addon) {
-    if (!hasDataPackage(this) && !hasOwnDataPackages(this)) {
+    if (!this._hasInitializedDependencyChecks) {
+      // we can't initialize in init because we need to super first
+      // but super calls shouldIncludeChildAddon which needs this
+      // so we initialize from the first call to shouldIncludeChildAddon
+      this._hasInitializedDependencyChecks = true;
+      this._excludeChildDataAddons = checkDependencyVersions(this.project);
```
igorT, line 27: Do we need/want a separate check for the inflector?

src/debug-macros.js
```diff
@@ -56,11 +56,16 @@ function getFlags(app, isDevelopingAddon) {
...
   }
-  const projectHelper = buildProjectHelper(app.project);
+  const checker = new VersionChecker(app);
   let isProd = process.env.EMBER_ENV === 'production';
-  let dataPackage = projectHelper.getAddon('ember-data');
-  let storePackage = projectHelper.getAddon('@ember-data/store');
-  let version = dataPackage ? dataPackage.pkg.version : storePackage.pkg.version;
+  let dataPackage = checker.for('ember-data');
+  let storePackage = checker.for('@ember-data/store');
+  let version = dataPackage.exists() ? dataPackage.version : storePackage.version;
```
igorT, line 63: Oh was this broken before? Nice catch

src/dependency-checks.js
```diff
@@ -0,0 +1,265 @@
...
+    versions[name] = require(`${name}/package.json`).version;
+  });
+
+  return versions;
+}
+
+/**
+ * Throws errors if invalid ember-data or ember-m3 configurations are detected.
+ * Returns a boolean `true` if the consuming app brings it's own versions of
+ * ember-data or ember-data packages. `false` otherwise.
```
igorT, line 246: The comment doens't mention ember-inflector but at least some checks here check for it. Great comment otherwise though 👍

src/dependency-checks.js
```diff
@@ -0,0 +1,265 @@
...
+  '@ember-data/store',
+];
+const minVersionForUsingOwnPackages = '3.14.0';
+
+class VersionChecker {
+  constructor(project) {
+    this._project = project;
+  }
+
+  *allAddons() {
```
igorT, line 24: Fancy

src/dependency-checks.js
```diff
@@ -0,0 +1,265 @@
...
+    let cache = this._project.packageInfoCache.entries;
+    let keys = Object.keys(cache);
+    for (let key of keys) {
+      let packageInfo = cache[key];
+
+      if (
+        packageInfo.pkg &&
```
igorT, line 31: Is `packageInfo.pkg` sometimes not there?

src/dependency-checks.js
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    return addons;
+  }
+  static forProject(project) {
+    return new VersionChecker(project);
+  }
+}
+
+function extractVersionsFromDeps(deps, m3Version, dataInfo) {
```
igorT, line 60: This is a pretty complex function, can you please add a comment explaining what it does. Maybe an example input -> output

package.json
runspired: yes

src/dependency-checks.js
```diff
@@ -0,0 +1,265 @@
...
+      seenVersions[pkgVersion]++;
+    }
+
+    return !seen;
+  });
+  return { versions: seenVersions, uniqueDeps };
+}
+
+/**
+ * Return
```
igorT, line 113: Is this on purpose?

index.js
runspired: yes

src/debug-macros.js
runspired: it was, so fixed :)

src/dependency-checks.js
runspired: yes, for things that are scoped the outer scope gets an entry (like `@babel`)

src/dependency-checks.js
runspired:
```suggestion
```

package.json
runspired: The previous node-tests existed for the previous dependency-management solution which has been removed.

index.js
runspired: apps won't see these logs though unless they run with `DEBUG=*` or `DEBUG=ember-m3`

index.js
runspired: I don't think we need to separate it out, and hopefully soon we can just drop the check altogether.

src/dependency-checks.js
runspired: Not trying to be, but wanted to match ember-cli-version-checker's current API so we can PR this update (the existing implementation fails to catch most addons because until addons complete initialization the thing it iterates isn't populated)

src/dependency-checks.js
runspired: Added

src/dependency-checks.js
runspired: Updated the comments

src/dependency-checks.js
runspired: Updated

stefanpenner:
`Dependency Management` tests appear to be logging errors, although appear to be green.

```
⚜️	🟡  [EXECUTING ...]	cd /home/runner/work/ember-m3/ember-m3/dependency-tests/tests/apps-with-errors/app-with-data-with-addon-with-data-v1 && npm install{ error:
   TypeError: process.stdout.clearLine is not a function
       at resolve (/home/runner/work/ember-m3/ember-m3/dependency-tests/utils.js:7:20)
       at new Promise (<anonymous>)
       at clearLine (/home/runner/work/ember-m3/ember-m3/dependency-tests/utils.js:6:10)
       at clearLines (/home/runner/work/ember-m3/ember-m3/dependency-tests/utils.js:27:9)
       at exec (/home/runner/work/ember-m3/ember-m3/dependency-tests/utils.js:39:11)
       at process._tickCallback (internal/process/next_tick.js:68:7)
       at Function.Module.runMain (internal/modules/cjs/loader.js:834:11)
       at startup (internal/bootstrap/node.js:283:19)
       at bootstrapNodeJSCore (internal/bootstrap/node.js:623:3) }
```

stefanpenner:
> Overwhelmingly the "files added" here are due to having added 27 new addons/apps for testing.

If this the path forward, what do we maintain the 27 generated addons/apps now in this repository going forward?

runspired:
> how do we maintain the 27 generated addons/apps now in this repository going forward? I am concerned this will introduce substantial friction and maintenance.

We can replace them at some point with something that creates them dynamically; however, doing so is non-trivial.

> Dependency Management tests appear to be logging errors, although appear to be green.

I'll look into it. The tests are passing locally but the node version is different (12 vs 8) and likely that is why.

igorT:
@runspired can you add the comments about the version checker conforming to the outside api and that we don't want to change that and that it's going away please?

igorT:
https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L146
says the method is ignoring things, but when I read the code flow I don't see it ignoring things. Can you please clairify/update the comments.

igorT:
https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L308
Can we name this so it indicates that it returns a value. Something like `checkIfAppBringsEmberData` or similar

igorT:
Is there an underlying reason why https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L296 is a method but https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L6 is a value

igorT:
Can we do something like `dataInfo = {}` to make it clear that it is optional/hint at what the shape is https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L154

igorT:
Because we don't have TS can you please add a comment that describes what https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L154 returns

igorT:
Can we break add the newline before rather than after here, https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L315
I got a bit confused with the grouping when trying to understand it

igorT:
Are we asserting here that there isn't more than one m3 present? We don't seem to need this for return values, can you please add a comment describing what the goal is https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L322

igorT:
https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L85
The first time you run through this method, I don't think we have asserted that data is singleton, no?

igorT:
It might be useful to break these blocks into named functions for readability https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L94

igorT:
https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L103
Does this account for the case when m3Version and fullED are both null?

igorT:
This method doesn't do any checking itself right? https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L154
Maybe we should rename?

igorT:
Maybe call this method assertHasSingleVersionIfPresent https://github.com/hjdivad/ember-m3/blob/d442f501009dd89ea4e7a5cfb91e0d6a121c0a7a/src/dependency-checks.js#L285

index.js
```diff
@@ -2,104 +2,42 @@
...
+    }
+
+    if (this._excludeChildDataAddons !== true) {
```
igorT, line 30: It would be nice to the include/exclude to be consistent so you don't have to juggle the nots. The method here is `shouldIncludeAddon`, can we have this prop be `shouldIncludeChildData`?

package.json
```diff
@@ -65,21 +65,26 @@
   "dependencies": {
     "babel-plugin-debug-macros": "^0.3.3",
     "broccoli-funnel": "^3.0.1",
+    "debug": "^4.1.1",
     "ember-cli-babel": "^7.19.0",
+    "ember-cli-version-checker": "^5.0.2",
     "esm": "^3.2.25",
-    "semver": "^7.3.2"
+    "semver": "^7.3.2",
+    "@ember-data/debug": "^3.17.0",
```
igorT, line 73: Should we default to the LTS version?

.github/workflows/ci.yml
```diff
@@ -69,6 +69,20 @@ jobs:
       - name: test
         run: yarn test
 
+  test-dependency-management:
+    name: 'Dependency Management' # rely on matrix for most of name
```
igorT, line 73: Not quite sure what `for most of name` means

runspired:
closing in favor of #695

PR #693, opened 2020-04-30
Dropping support for < ember-data 3.12

hjdivad:
This will be a major version bump (2.0.0)
This aims to drop unused branching, improving hygiene and reducing complexity.

- [x] un-wip

### DIFF
--- a/addon/initializers/m3-store.js
+++ b/addon/initializers/m3-store.js
@@ -1,5 +1,4 @@
 import '../mixins/store';
-import { HAS_EMBER_DATA_PACKAGE } from 'ember-m3/-infra/packages';
 import { DEBUG } from '@glimmer/env';
 import require from 'require';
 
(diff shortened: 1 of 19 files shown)


addon/initializers/m3-store.js
```diff
@@ -20,10 +19,13 @@ function initializeDebugAdapter(registry) {
...
 
 export function initialize(application) {
-  if (!HAS_EMBER_DATA_PACKAGE) {
-    application.inject('route', 'store', 'service:store');
-    application.inject('controller', 'store', 'service:store');
-  }
+  // This should be unnecessary
+  // it is done by the meta package
+  // but it should be done by the store package
+  // https://github.com/emberjs/data/issues/7158
```
runspired, line 25: responded there, but TL;DR is we don't expect the store package to do this, the individual packages present a modernized version of EmberData as much as possible. This injection is a legacy piece we are phasing out. Apps should migrate away.

addon/mixins/store.js
```diff
@@ -276,21 +264,14 @@ function createRecordDataFor(modelName, id, clientId, storeWrapper) {
     );
   }
 
+  // TODO: what is the purpose of this check?
   if (HAS_EMBER_DATA_PACKAGE) {
```
runspired, line 268: probably because there was nothing to super to if you don't have the meta package, although that is something we're going to fix. TL;DR a mistake (maybe rectified? possibly forgotten during my leave?) was made wherein the meta package is what configured the store to use @ember-data/model.

PR #695, opened 2020-05-01
utilize ember-data's project trim

hjdivad:
Due to the complexity with managing highlander dependencies, and the potential difficulties that our consumers may have attempting to resolve those issues. We are proposing an alternative approach, one that divorces utilizing trim from ember-m3 being a "meta package".

Specifically, we now expect either a user to simply use ember-data, or if they are power-users wanted to reduce asset size, they can instead use the individual data packages, from the root of their application. We will include a list in the readme, of known compatible versions and ember-m3 now throws a helpful error message if a violation occurs.

This keeps the complexity down, the only penalty remaining is that power-user engineers opting into trim 2-3 additional node_modules (`@ember-data/model` `@ember-data/store` `ember-inflector`)

---

Remaining Tasks
- [x] update README to explain dependencies + install w/ and w/o trim
- [x] tests
- [x] ~stretch: drop esm~ dropped in #705


index.js
```diff
@@ -2,105 +2,77 @@
...
-    if (addon.name.startsWith('@ember-data') || addon.name === 'ember-inflector') {
-      /*
-      console.log(
-        `⚠️  ember-m3 is excluding ${addon.name} version ${addon.pkg.version} from the build`
-      );
-      */
-      return false;
-    }
-    return true;
+    debugger;
```
stefanpenner, line 72: debugger

index.js
```diff
@@ -2,105 +2,77 @@
...
-  let addons = addon.project.addonPackages;
-  return addons['ember-data'] !== undefined;
-}
-function hasOwnDataPackages(addon) {
-  let addons = addon.project.addonPackages;
 
-  for (let i = 0; i < allDataPackages.length; i++) {
-    if (addons[allDataPackages[i]] !== undefined) {
-      return true;
+function assertValidEmberData(addon) {
```
stefanpenner, line 15: We should likely generalize this, and move this or its constituent parts to ember-cli-version-checker and thoroughly test it.

index.js
```diff
@@ -2,105 +2,82 @@
...
+  let emberDataFound = false;
+  let emberDataVersionsFound = [];
+  let emberDataModelFound = false;
+  let emberDataModelVersionsFound = [];
+  let emberDataStoreFound = false;
+  let emberDataStoreVersionsFound = [];
+  let emberInflectorFound = false;
+  let emberInflectorVersionsFound = [];
+
+  for (let { name, pkg } of checker.allAddons()) {
```
runspired, line 19: we'll need to either fix the project wide check in ember-cli-version-checker or bring the replacement from the other PR over to this one.

addon/-infra/potential-packages.js
```diff
@@ -1,18 +0,0 @@
-/**
```
runspired, line 1: seems odd to be deleting this as part of this PR

hjdivad: checking the CI you're right this is why the 3.12 tests are failing

hjdivad:
see https://github.com/ember-cli/ember-cli-version-checker/pull/206

PR #705, opened 2020-05-05
cleanup: BREAKING drop EmberData<3.16 support for 2.0

runspired:


src/debug-macros.js
```diff
@@ -56,26 +37,13 @@ function getFlags(app, isDevelopingAddon) {
...
-  } catch (e) {
-    try {
-      features = app.project.require('@ember-data/-build-infra/src/features');
-    } catch (e) {
-      features = requireEsm('../addon/-infra/potential-features.js').default;
-    }
-  }
+
+  // >= 3.15.0.beta
+  features = app.project.require('@ember-data/private-build-infra/src/features')(isProd);
```
hjdivad, line 46: @runspired what's the plan for making this public API?

Requiring addons to `require` `private-build-infra` to have features that depend on ED feature flags seems bad.

hjdivad:
@runspired LGTM to merge after you resolve the conflicts.

Requiring `private-build-infra` is obviously a smell and addons needing to do this is something to be fixed in ember data but no reason to block this pr

src/debug-macros.js
runspired: @hjdivad there is no plan, but that's something we can maybe chat about at the F2F

PR #719, opened 2020-05-18
updates readme with tests/dummy example and fix dummy

betocantu93:
I had a hard time with hello world, the documented example just doesn't work. The tests/dummy does work, so this might save some time for someone


hjdivad:
Thanks very much @betocantu93!  Love to see contributions like this 🎉

PR #720, opened 2020-05-18
Add Alberto to contributors

hjdivad:


hjdivad:
@betocantu93 LMK if you'd prefer to not be added.  Either way thanks again for #719.  Patches always welcome. ^_^

betocantu93:
no problem @hjdivad, we have some serious use case for this kind of flexibility, so I'll be playing around and maybe update some docs while doing so

PR #763, opened 2020-06-13
fix: update should trigger flush without included

runspired:
This uses an intimate API (`store.didSaveRecord`) to capture the return of updates and schedule a flush of the changes.

The schedule is only because of our desire to batch these notifications with any generated from `included` records, if we don't care about the batching we can just call flushChanges straight up and it would affect only the primary data while a separate flush would occur for the changes from `included`.

I explored two other alternatives that achieve this, both have downsides.

1) flush within `RecordData.didCommit`. This only works if we use `run.next` or similar to defer the flush until after the changes are notified (notifications are generated based on the returned changedKeys generated by `didCommit`). It requires accessing the `StoreWrapper._store` to get the weak map key as well.

2) chaining the promise within the resolver given to the private `store.scheduleSave` method to perform the flush. This felt more brittle.

Regardless of approach taken here (including closing without merging) we should work to ensure this is possible via public API in EmberData.

### DIFF
--- a/addon/services/store.js
+++ b/addon/services/store.js
@@ -268,4 +268,15 @@ export default class M3Store extends Store {
       return internalModel;
     }
   }
+
+  // this private hook is used for update/create/delete record responses which don't
+  // take the _push codepath
+  didSaveRecord() {
+    super.didSaveRecord(...arguments);
+    // EmberData wraps the call to didSaveRecord in a run.join of
+    // it's internal backburner, and pushes sideloaded data from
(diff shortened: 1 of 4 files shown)


hjdivad:
> Regardless of approach taken here (including closing without merging) we should work to ensure this is possible via public API in EmberData.

yep strong +1

also tests are red although interestingly only with FF on.  Seems a bit odd that custom model classes would affect batching like this.

PR #770, opened 2020-06-15
cleanup: Cleanup and simplify RecordArray code

igorT:
RecordArrays had leftover code from the array proxy days, which this
commit cleans up. Also adds feature flags in places where they were
missing

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -549,9 +549,11 @@ export default class M3RecordData {
   }
 
   removeFromRecordArrays() {
-    this._recordArrays.forEach(recordArray => {
-      recordArray._removeRecordData(this);
-    });
+    if (CUSTOM_MODEL_CLASS) {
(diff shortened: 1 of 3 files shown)


addon/record-data.js
```diff
@@ -549,9 +549,11 @@ export default class M3RecordData {
...
 
   removeFromRecordArrays() {
-    this._recordArrays.forEach(recordArray => {
-      recordArray._removeRecordData(this);
-    });
+    if (CUSTOM_MODEL_CLASS) {
+      this._recordArrays.forEach(recordArray => {
+        recordArray._removeRecordData(this);
+      });
+    }
```
hjdivad, line 556: Why do we only call this when `CUSTOM_MODEL_CLASS` is on?

If the answer is b/c `removeFromRecordArrays` is only called under such circumstances we should assert that

igorT: :+1: will add the assertion

PR #771, opened 2020-06-15
cleanup: Fix array iteration off by one

igorT:
_removeInternalModels loop had a off by one error that was hidden by a
superfluous call to `isResolvedValue`.

`isResolvedValue` is pointless in this place, as in a tracked array
all of the models are always resolved

### DIFF
--- a/addon/m3-tracked-array.js
+++ b/addon/m3-tracked-array.js
@@ -102,17 +102,15 @@ export default class M3TrackedArray extends EmberObject.extend(MutableArray) {
     if (CUSTOM_MODEL_CLASS) {
       throw new Error('Should not be calling _removeInternalModels when CUSTOM_MODEL_CLASS is on');
     } else {
-      for (let i = this.content.length; i >= 0; --i) {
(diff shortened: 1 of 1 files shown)


addon/m3-tracked-array.js
```diff
@@ -102,17 +102,15 @@ export default class M3TrackedArray extends EmberObject.extend(MutableArray) {
     if (CUSTOM_MODEL_CLASS) {
       throw new Error('Should not be calling _removeInternalModels when CUSTOM_MODEL_CLASS is on');
     } else {
-      for (let i = this.content.length; i >= 0; --i) {
```
runspired, line 105: so in the past we iterated an extra item here (one too many)

hjdivad:
```
isResolvedValue is pointless in this place, as in a tracked array
all of the models are always resolved
```

not when you mutate the array

hjdivad:
> ```
> isResolvedValue is pointless in this place, as in a tracked array
> all of the models are always resolved
> ```
> 
> not when you mutate the array

welp I guess we always eagerly resolve on mutation so seems fine

PR #774, opened 2020-06-17
[cleanup] M3TrackedArray extends M3RecordArray

igorT:


tests/unit/model/tracked-array-test.js
```diff
@@ -190,16 +190,12 @@ module('unit/model/tracked-array', function(hooks) {
...
-      willChange: (array, index, removeCount) => {
-        observerWillChange = removeCount;
-      },
       didChange: (array, index, removeCount) => {
         observerDidChange = removeCount;
       },
     });
 
     run(() => chapter2.unloadRecord());
-    assert.equal(observerWillChange, 1, 'removal of the record notified to be removed');
```
runspired, line 202: do we no longer fire willChange?

addon/record-array.js
```diff
@@ -33,7 +37,9 @@ if (CUSTOM_MODEL_CLASS) {
     init() {
       super.init(...arguments);
       this._references = [];
-      this._objects = A();
+      if (!this._objects) {
```
runspired, line 40: can we obscure this property even more to prevent access? would be nice to kill off these _ props by moving them to a weakmap or some such

tests/unit/model/tracked-array-test.js
igorT: Because we didn't fire it for recordArrays, and did for tracked, so now settling on not firing them

addon/record-array.js
igorT: Given the _ and the non similarity to public apis, I don't think it's as bad as `content`. It's used in child classes, so having it on the array is nice.

Issue #782, opened 2020-06-23
Make the params option for queryURL more explicit


2hu12:
The current `params` option has different effect for `GET` and `POST` which is like the `data` option for `jQuery.ajax`.

But it will be inconvenient for `POST` method url with query parameters, this is common for some api design like “rest.li”.

Could we split this `params` into two options, for example, `queryParams` and `payload` or some more idiomatic option names? For `GET` it can only have `queryParams` and for `POST` it can have both.

2hu12:
I'd like to take this task, but I'm kinda uncertain about where should we place this modification, `ember-m3/query-cache` or `RESTAdapter.ajax` method. I think I do it in `ember-m3` at first.

hjdivad:
@2hu12 if you have time to contribute that would be great.  I think we'd want to do this in `query-cache`, either in or before we get to `_adapterQueryURL`.

Not sure what the best API is but maybe something like `params` and `urlParams`, where the latter are added to the URL and the former are treated as they currently are (where we hand them off to either `adapter.queryURL` or `adapter.ajax`)

hjdivad:
fixed in #784

PR #784, opened 2020-06-25
Add queryParams option for Store.queryURL

2hu12:
FYI #782

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,11 @@ normalization to JSON API.
   - `options.params` defaults to `null`. The parameters to include, either in the URL (for `GET`
     requests) or request body (for others).
 
+  - `options.queryParams` defaults to `null`. The parameters that will be
+    converted to query string and appended to the URL, especially useful for
+    `POST` method which needs both URL with query string(`options.queryParams`)
+    and payload data(`options.params`).
+
   - `options.cacheKey` defaults to `null`. A string to uniquely identify this
     request. `null` or `undefined` indicates the result should not be cached.
     It is passed to `serializer.normalizeResponse` as the `id` parameter, but
--- a/addon/query-cache.js
+++ b/addon/query-cache.js
@@ -6,6 +6,7 @@ import { assign } from '@ember/polyfills';
 import MegamorphicModel from './model';
 import M3QueryArray from './query-array';
 import { CUSTOM_MODEL_CLASS } from 'ember-m3/-infra/features';
+import serializeQueryParams from './utils/serialize-query-params';
 
 function stripSlash(str, stripLeading, stripTrailing) {
   let startSlash = stripLeading && str.charAt(0) === '/';
@@ -27,6 +28,7 @@ export default class QueryCache {
     url,
     {
       params = null,
+      queryParams = null,
       method = 'GET',
       cacheKey = null,
       reload = false,
@@ -44,7 +46,7 @@ export default class QueryCache {
     }
 
     let cachedPromise = cacheKey ? this._queryCache[cacheKey] : undefined;
-    let adapterUrl = this._buildUrl(url);
+    let adapterUrl = this._buildUrl(url, queryParams);
     let loadPromise;
 
     if (backgroundReload || reload || cachedPromise === undefined) {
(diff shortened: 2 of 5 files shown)


addon/query-cache.js
```diff
@@ -2,6 +2,7 @@ import { Promise as RSVPPromise } from 'rsvp';
 import { assert } from '@ember/debug';
 import { get } from '@ember/object';
 import { assign } from '@ember/polyfills';
+import { serializeQueryParams } from '@ember-data/adapter/-private';
```
2hu12, line 5: Should we port this util function instead of adding a new dependency?

hjdivad: yes

1. we definitely don't want a dependency on `@ember-data/adapter`
2. this is also the reason the tests are failing against latest ember-data https://github.com/hjdivad/ember-m3/pull/784/checks?check_run_id=808449585#step:5:37

README.md
```diff
@@ -194,6 +194,11 @@ normalization to JSON API.
   - `options.params` defaults to `null`. The parameters to include, either in the URL (for `GET`
     requests) or request body (for others).
 
+  - `options.urlParams` defaults to `null`. The parameters that will be
```
rwjblue, line 197: Why `urlParams` here vs something like `queryParams`? Using `urlParams` seems somewhat odd to me because (at least for `GET` requests) `options.params` is _also_ put in the URL.

addon/utils/serialize-query-params.js
```diff
@@ -0,0 +1,65 @@
...
+        buildParams(key, obj[key]);
+      }
+    }
+    return s;
+  }
+
+  return buildParams('', queryParamsObject)
+    .join('&')
+    .replace(/%20/g, '+');
+}
```
hjdivad, line 49: Thanks for porting this over.  It's a lot clearer now and we don't have the dep.

Looking at the specific implementation though, I think we should use a much simpler version that only handles primitive values (so no array handling or object handling).

Users can still use those values, but they would have to put them in the URL themselves, or encode them as strings before passing them to `urlParams`.

The reason I don't want this implementation in particular is that there isn't a standard encoding scheme for arrays and nested structures to a URL, and the right encoding scheme will depend on the API.  So I don't think we should use any particular encoding, beyond stuff that will work with more or less all APIs (i.e. simple `${key}=${value}`).

So this means that things in params that wind up in the content get `toJSON` and things that wind up in the URL as query params get `toString` which seems like a reasonable mental model for users IMHO.

README.md
2hu12: Kinda agree with you, the `urlParams` indeed does the same thing as `params` for `GET` requests. The initial purpose of #782 is to split the `params` to `urlParams` which will be appended to the URL as the query string and body `payload`. The current solution will cover this purpose, but it is sort of redundant, I think that @hjdivad suggests this way mainly for the backward compatiblity.

I've thought about some alternatives, e.g.
- Remove the `params` option, add `urlParams` and `payload`.
- For `GET` requests, it could only has one of `urlParams` or `params`, or an error will be thrown. Or even further, it could only use `urlParams` for `GET` requests.

hjdivad: @2hu12 I think @rwjblue was mainly suggesting a different name rather than a different implementation.  I agree that `queryParams` is a more typical name than `urlParams`.  I should have thought a bit more before making that suggestion, sorry.

You're right that I was thinking about backwards compatibility.  But maybe the better approach is to do your original suggestion of `queryParams` and `payload` / `body` and keep `options.params` but deprecated.  What do you think?

I really don't want to impose restrictions based on the request method because not having a request body for `GET` requests is just a convention that servers need not honour.

2hu12: @hjdivad Thanks for the explanation.

I think my original purpose( add `queryParams` and `payload` options and deprecate `params` option) won't be implemented only in `ember-m3`, for that the `params` is actually used by the `RESTAdapter.ajax` and it will be treated differently for different request methods, which is also related to the `GET` payload restrictions, for example, even if we want to provide the `payload` option for one `GET` request, it is impossible currently.

I just update the `urlParams` to `queryParams` for now, and maybe if one day `RESTAdapter.ajax` add options for `queryParams` and `payload`, we could follow it.

hjdivad:
Thanks @2hu12!

PR #787, opened 2020-06-26
replace computeNestedModel and computeAttributeReference with computeAttribute

igorT:


addon/record-data.js
```diff
@@ -42,6 +43,16 @@ class M3SchemaInterface {
     this._suppressNotifications = false;
   }
 
+  nested(object) {
+    schemaTypesInfo.set(object, NESTED);
+    return object;
+  }
```
hjdivad, line 49: comment this public api

addon/record-data.js
```diff
@@ -42,6 +43,16 @@ class M3SchemaInterface {
...
+
+  reference(object) {
+    schemaTypesInfo.set(object, REFERENCE);
+    return object;
+  }
```
hjdivad, line 54: comment this public api

addon/record-data.js
hjdivad: also add docs in readme

addon/resolve-attribute-util.js
```diff
@@ -114,13 +116,32 @@ export function resolveValue(key, value, modelName, store, schema, record, paren
   const schemaInterface = recordData.schemaInterface;
 
   // First check to see if given value is either a reference or an array of references
```
hjdivad, line 118: this comment is now out of place

addon/resolve-attribute-util.js
```diff
@@ -114,13 +116,32 @@ export function resolveValue(key, value, modelName, store, schema, record, paren
...
-  let reference = computeAttributeReference(key, value, modelName, schemaInterface, schema);
-  if (reference !== undefined && reference !== null) {
+  let computedValue, nested, reference;
+  if (schema.useComputeAttribute()) {
```
hjdivad, line 120: this is a hot path; at a minimum we will want to cache this

addon/resolve-attribute-util.js
```diff
@@ -114,13 +116,32 @@ export function resolveValue(key, value, modelName, store, schema, record, paren
...
+    computedValue = computeAttribute(key, value, modelName, schemaInterface, schema);
+    if (computedValue && typeof computedValue === 'object') {
```
hjdivad, line 122: should confirm that this is the best way to check for POJOness in a hot path

addon/services/m3-schema-manager.js
```diff
@@ -21,7 +23,16 @@ export default class SchemaManager extends Service {
...
+    let schema = this.get('schema');
+    if (this.useComputeAttribute()) {
+      let computedValue = schema.computeAttribute(key, value, modelName, schemaInterface);
+      let valueType = schemaTypesInfo.get(computedValue);
+      if (valueType === REFERENCE) {
+        return computedValue;
+      }
+    } else {
+      return schema.computeAttributeReference(key, value, modelName, schemaInterface);
+    }
```
hjdivad, line 35: I would not expect someone opted in to the new API to end up with call paths going through the old API.  Is there a good reason for this?

addon/services/m3-schema-manager.js
```diff
@@ -35,7 +46,30 @@ export default class SchemaManager extends Service {
...
+    if (this.useComputeAttribute()) {
+      let computedValue = schema.computeAttribute(key, value, modelName, schemaInterface);
+      let valueType = schemaTypesInfo.get(computedValue);
+      if (valueType === NESTED) {
+        return computedValue;
+      }
+    } else {
+      return schema.computeNestedModel(key, value, modelName, schemaInterface);
+    }
+  }
```
hjdivad, line 59: I would not expect someone opted in to the new API to end up with call paths going through the old API.  Is there a good reason for this?

addon/services/m3-schema-manager.js
```diff
@@ -35,7 +46,30 @@ export default class SchemaManager extends Service {
...
+
+  useComputeAttribute() {
+    let schema = this.get('schema');
+    let isComputeAttributeReferenceMissing = !schema.computeAttributeReference;
+    let defaultPrototype = DefaultSchema.prototype;
+    let usingDefaultSchemaMethods =
+      defaultPrototype.computeAttributeReference === schema.computeAttributeReference &&
+      defaultPrototype.computeNestedModel === schema.computeNestedModel;
+    return isComputeAttributeReferenceMissing || usingDefaultSchemaMethods;
```
hjdivad, line 72: At a minimum we'll need to cache this

hjdivad, line 72: It makes more sense to me to check against `computeAttribute` being implemented than the inverse

package.json
```diff
@@ -65,7 +65,7 @@
   "dependencies": {
     "babel-plugin-debug-macros": "^0.3.3",
     "broccoli-funnel": "^3.0.3",
-    "ember-cli-babel": "^7.21.0",
+    "ember-cli-babel": "7.18",
```
hjdivad, line 68: This seems suspect.  We should probably remove it.

addon/resolve-attribute-util.js
igorT: :+1: cached in a weak map

addon/resolve-attribute-util.js
igorT: good catch, I asked and it turned out that checking for null explicitly is better, fixed

addon/services/m3-schema-manager.js
igorT: Good point, fixed

addon/services/m3-schema-manager.js
igorT: Cached using a weakmap

igorT: Fixed

package.json
igorT: Yup, it was leftover from testing, as linking was unhappy with babel issues at some point

tests/unit/model/changed-attrs-test.js
```diff
@@ -1284,6 +1319,7 @@ module('unit/model/changed-attrs', function(hooks) {
     );
   });
 
+  /*
```
igorT, line 1322: Test not valid for the new world, will bring back in the old tests

addon/resolve-attribute-util.js
igorT: fixed

addon/services/m3-schema-manager.js
igorT: Good point, fixed

addon/record-data.js
igorT: commented

addon/record-data.js
igorT: done

tests/unit/model/state-test.js
```diff
@@ -6,138 +6,193 @@ import { isArray } from '@ember/array';
...
-        } else if (Array.isArray(refValue)) {
-          return refValue.map(x => ({
-            type: null,
-            id: x,
-          }));
-        }
-        return null;
-      }
+let computeNestedModel = function computeNestedModel(key, value) {
+  if (value && typeof value === 'object' && value.constructor !== Date && !isArray(value)) {
```
runspired, line 10: I believe we can clean up this check by removing the Date check (seems no date usage below).

tests/unit/model/tracked-array-test.js
```diff
@@ -7,297 +7,336 @@ import DefaultSchema from 'ember-m3/services/m3-schema';
...
-        }
-
-        computeNestedModel(key, value /*, modelName, schemaInterface */) {
-          if (typeof value === 'object' && value !== null && !Array.isArray(value)) {
-            return {
-              attributes: value,
-            };
-          }
+function computeNestedModel(key, value /*, modelName, schemaInterface */) {
+  if (typeof value === 'object' && value !== null && !Array.isArray(value)) {
```
runspired, line 11: we do effectively this check a lot of places in both our code and our tests, and it seems slightly different every time. This is likely to be a hot path in our actual code, so we should audit at some point to ensure that the simplest bailout is being done prior to the typeof check.

tests/unit/projection-test.js
```diff
@@ -1296,1828 +1404,1830 @@ module('unit/projection', function(hooks) {
...
-          data: {
-            id: BOOK_ID,
-            type: BOOK_EXCERPT_PROJECTION_CLASS_PATH,
-            attributes: {
-              title: BOOK_TITLE_2,
-            },
-          },
-        });
-      });
+  module('creating/updating projections with resolutions', function(hooks) {
```
runspired, line 3057: lets split this module out and make sure we test the old hooks (removed accidentally)

tests/unit/model/state-test.js
igorT: Cleaned up

tests/unit/projection-test.js
igorT: done

igorT:
@hjdivad addressed all of the feedback, and @runspired gave it a read through, so will be merging

PR #788, opened 2020-06-29
TrackedArray and ReferenceArray inherit from same base class with ff off

igorT:
I was hoping to not have to do this, as we will be dropping this code shortly
anyways, but it was blocking future refactors, and wasn't too much work


hjdivad:
@igorT following up on our conversation OOB, let's rename this to something like `ManagedArray`.  I think the name of this class is much more important than the name of the base class or of the query array, because this is the class users are most likely to see and interact with, and I'd like the name to be descriptive to them rather than focusing on internal concerns.

PR #810, opened 2020-07-24
easier transformValue overriding

betocantu93:
Easier to extend the transform functionality without much hassle, In my particular use case I had to send more params to `this._schema.transformValue`, like `_topModel`

The reality is that I have structured schema with the type information for each attr, a [json schema](https://json-schema.org/) so each Megamorphic topModel have a json schema, so all the transforms and default values must be read from this schema per access, so my method call reads like this `this._schema.transformValue(this._modelName, key, value, this._topModel.schemaVersion)` and probably I would also favor `this._schema.getDefaultValue(this._modelName, key, this._topModel.schemaVersion)` 

```ts
_getDefaultValue(modelName, key){
  return this._schema.getDefaultValue(modelName, key)
}
_transformValue(modelName, key, value){
  return this._schema.transformValue(modelName, key, value)
}
```
So now I can easily do
```ts

MegamorphicModel.reopen({
  schemaVersion: computed('_topModel', function() {
    return this.get('_topModel.schema_version_id');
  }),

  _getDefaultValue(modelName, key) {
    return this._schema.getDefaultValue(modelName, key, this.schemaVersion);
  },

  _transformValue(modelName, key, rawValue) {
    return this._schema.transformValue(modelName, key, rawValue, this.schemaVersion);
  }
});

```

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -415,7 +415,8 @@ export default class MegamorphicModel extends EmberObject {
       }
     }
 
-    let value = this._schema.transformValue(this._modelName, key, rawValue);
+    let value = this._transformValue(this._modelName, key, rawValue);
+
     return (this._cache[key] = resolveValue(
       key,
       value,
@@ -424,6 +425,10 @@ export default class MegamorphicModel extends EmberObject {
       this._schema,
       this
     ));
+  }
+
+  _transformValue(modelName, key, rawValue) {
+    return this._schema.transformValue(modelName, key, rawValue);
   }
 
   get id() {
--- a/addon/model.js
+++ b/addon/model.js
@@ -432,7 +432,7 @@ export default class MegamorphicModel extends EmberObject {
   }
 
   _transformValue(modelName, key, rawValue) {
-    return this._schema.transformValue(modelName, key, rawValue, this._topModel);
+    return this._schema.transformValue(modelName, key, rawValue);
   }
 
   get id() {
--- a/addon/model.js
+++ b/addon/model.js
@@ -407,7 +407,7 @@ export default class MegamorphicModel extends EmberObject {
         return get(this, key);
       }
 
-      let defaultValue = this._schema.getDefaultValue(this._modelName, key);
+      let defaultValue = this._getDefaultValue(this._modelName, key);
 
       // If default value is not defined, resolve the key for reference
       if (defaultValue !== undefined) {
@@ -427,8 +427,12 @@ export default class MegamorphicModel extends EmberObject {
     ));
   }
 
+  _getDefaultValue(modelName, key) {
+    return this._schema.getDefaultValue(modelName, key);
+  }
+
   _transformValue(modelName, key, rawValue) {
-    return this._schema.transformValue(modelName, key, rawValue);
+    return this._schema.transformValue(modelName, key, rawValue, this._topModel);
   }
 
   get id() {


betocantu93:
Update: I guess I didn't saw this coming, but sending the topModel and reading from it would cause a stack overflow... closing for now, while I understand how to better handle this...

betocantu93:
Update: I did end up using this but also added a transform blacklist, so some values won't be transformed thus ending the infinite recursion.

hjdivad:
@betocantu93 thanks for digging in to this.

The current PR implementation adds a couple of `_fn` extension points: we should stick to the convention of treating functions with a leading `_` as private.  Also I'd like to avoid having the model class itself be an extension point.

That said the problem you're raising makes total sense: you need more context than is currently provided to the `schema.models` section for default values and transforms.

There's also the second problem of there not being a public API for getting the top model of a nested model.

Both of these seem to me to be well worth solving.

---

For the first problem the best solution that I see is to deprecate specifying defaults and transforms on `schema.models` and instead have users do these actions within `computeAttribute`

This should work today already for defaults, although transforms are [current run eagerly](https://github.com/hjdivad/ember-m3/blob/2efe91131b79333b2e9245e44dd90dfa7347b58c/addon/model.js#L417-L418) this could easily be moved into the [else branch](https://github.com/hjdivad/ember-m3/blob/2efe91131b79333b2e9245e44dd90dfa7347b58c/addon/resolve-attribute-util.js#L135) for the old compute hooks.

@betocantu93 what do you think?

betocantu93:
@hjdivad 

> The current PR implementation adds a couple of `_fn` extension points: we should stick to the convention of treating functions with a leading `_` as private. Also I'd like to avoid having the model class itself be an extension point.

> That said the problem you're raising makes total sense: you need more context than is currently provided to the `schema.models` section for default values and transforms.
> 
> There's also the second problem of there not being a public API for getting the top model of a nested model.
> 
> Both of these seem to me to be well worth solving.
> 
> For the first problem the best solution that I see is to deprecate specifying defaults and transforms on `schema.models` and instead have users do these actions within `computeAttribute`

I agree, centralizing everything within computeAttribute makes sense to me, deprecating eager transform.
A couple of notes:

I also think in this truly dynamic world, computeAttribute should receive a little bit more context than just the rawValue, I don't know for if i'm biased by my use case, but maybe `_topModel` should be enough context within computeAttribute, but again, that's just to solve my requirement, "every _topModel belongsTo a DS.Model, which has more typed info"

betocantu93:
Also, I think `schema.models` transforms for known models and attrs works great, but for truly dynamism is not really enough... I had to do some tricks inside the serializer like flattening nested objects (to reduce the memory usage) and also to encode the keys to get some deep transforms going

```ts
{ 
  extras: {
    prop1: {
      pictures: ['url'],
      price: 200.50,
      timestamp: "2020/10/12"
    },
     prop2: {
      pictures: ['url'],
      price: 200.50,
      timestamp: "2020/10/12"
    }
}
```
to
```ts
{
  `extras::prop1::pictures[array]`: [],
  `extras::prop1::price[number]`: 200.50,
  `extras::prop1::timestamp[date]`: "2020/12/10",
  `extras::prop2::price[number]`: 200.50,
  `extras::prop2::pictures[array]`: [],
  `extras::prop2::timestamp[date]`: "2020/12/10"
}
```

Ignoring the flattening, I think maybe some kind of convention to apply transforms on dynamic keys using maybe using regex could great (I understand this is not as easy task or silver bullet).

So I think:
1. More context should be added to computeAttribute hook.
2. By deprecating `schema.models` transforms and defaults I see two options
   1.  Create a transformValue function on `m3-schema` which runs before the computeAttribute but with the new context and in more public fashion on `m3-schema` service
   2. Just show some patterns and example on how to apply transforms inside computeAttribute
3. I found useful to augment the current transformValue to receive an argument "type" which defaults to `deserialize` and also call this function inside the serialization process but with `serialize`, using the according ember-data transforms

hjdivad:
I'm very +1 on having a smaller API surface area via `computeAttribute`, and providing that function more context, especially indirectly via `schemaInterface`.

For top model access maybe something like
```
schemaInterface.topModel()
```

So then you'd

```
// m3-schema
{
  computeAttribute(key, value, modelName, schemaInterface) {
    let topm = schemaInterace.topModel();
    // ...
  }
}
```

If you can do everything in `computeAttribute`, then having a `schema.models` can serve as an example in documentation of a clean way to handle simple schemas, and `computeAttribute` the hook where you can add more sophistication.

@betocantu93 does 👍 make sense to you?

If you have cycles to work on this I'd be happy to review moving the API in this direction

betocantu93:
@hjdivad just listing my understanding 
1. Deprecate transformValue
2. Add more context to computeAttribute via schemaInterface
3. Add a clear example of how `schema.models` and transforming could work via computeAttribute
i.e
```ts
//m3-schema
{
  computeAttribute(key, value, modelName, schemaInterface) {
    value = myVeryOwnTransformsImplementation(key, value, modelName, schemaInterface.topModel(), get(this, `models.${modelName}.${key}.transforms`);
  }
}
```
4. Add backwards compatibility for old apis.

sounds good to me.

I'll get a draft PR going and see how it goes thanks for your feedback

hjdivad:
@betocantu93 sounds like a plan; looking forward to the pr ❤️

hjdivad:
@betocantu93 did you get a chance to work on [the pr](https://github.com/hjdivad/ember-m3/pull/810#issuecomment-674442242)

betocantu93:
Hey @hjdivad sorry for the extended silence/inactivity, unfortunately no, I've been pretty busy lately with a major product release, which includes releasing ember-m3 to production (so happy for this), but I do plan to tackle this soon! It's in the back of my head because it's literally on our core now, I'm going to try to get a PR going this weekend

hjdivad:
@betocantu93 no worries I totally understand.  Looking forward to the pr when you get some time to get back at it

Issue #807, opened 2020-07-21
Raw values stringified


betocantu93:
Hello, I'm working in a PoC using ember-m3 and DS.Models for a possible incremental migration, it's going great so far, so many unlocked possibilities! I'm using the current `master` branch for the recently added computeAttribute API, but I'm getting a weird error... sometimes arrays and/or objects raw values get stringified  and I'm unable to replicate it consistently yet, any suggestions? thanks in advance

![Screen Shot 2020-07-20 at 19 30 05](https://user-images.githubusercontent.com/9092644/88005072-10135e80-cace-11ea-9fb8-9f676398cc04.png)

Here is a reproduction

https://github.com/betocantu93/show-stringified-bug

If you open ember inspector/data tab select first ref:document, get the instance into $E and...

try: `$E.get('prop1.table)` or : `$E.get('prop1.chair)` where prop1 is a nested object

you will get the table array stringified or chair object stringified

I'm not sure why, but this happens with app reloads

betocantu93:
It seems that its always stringified unless you reload the app with the ember inspector/data tab opened with the faulty model type selected, like if its lazy evaluated or something, gets stringified?

betocantu93:
I think its a bug with the ember inspector, if you reload in the data tab with a **Megamorphic** type selected, it gets this weird bug.

betocantu93:
I just found out that if you kept a megamorphic type selected on the ember inspector _data_ tab between app reloads other megamorphic types get their arrays and object raw values stringified, I opened an issue here: https://github.com/hjdivad/ember-m3/issues/807

It's kind of weird, the literal workaround is to just don't keep a megamorphic type selected between reloads...

SYU15:
@betocantu93 thanks for flagging this. I can't repro on the m3 dummy app however, I am fairly confident the issue is probably related to [these lines of code](https://github.com/hjdivad/ember-m3/blob/master/addon/adapters/m3-debug-adapter.js#L64-L68) where we stringify the object value so that it doesn't render as `[object object]` in the inspector. We put this in place due to the constraints of Ember Inspector which does not traverse the data structures deep enough for performance reasons iirc to display in the inspector table.

So this is ultimately in place because the table view as it is now is not the ideal way of displaying this type of data and we need to put a workaround in place.

So I guess there are 2 options (both are possible):
[...]

Issue #769, opened 2020-06-15
changed attributes issues


hjdivad:
https://github.com/pzuraq/m3-changed-attributes-bug

hjdivad:
It looks like Ember Data's `CUSTOM_MODEL_CLASS` support does not enable custom models to have additional features in changed attributes.  @igorT looks like some work is needed on `CUSTOM_MODEL_CLASS`?

---

There's a workaround available which is to call `changedAttributes on the record rather than the snapshot, e.g.

```
// app/adapters/application.js

export default class Adapter {
  updateRecord(store, modelClass, snapshot) {
    // as opposed to snapshot.changedAttributes()
    changedAttrs = snapshot.record.changedAttributes();
  }
}
```

hjdivad:
@igorT any updates?

PR #817, opened 2020-08-02
temporary stringify fix

betocantu93:
Temporary fix for #807 

The bug happens when iterating over _recordValues and find a reference to another [_MegamorphicModel_](https://github.com/hjdivad/ember-m3/blob/971f3764b1aa7121e34ade4df82eaea8026d8b75/addon/adapters/m3-debug-adapter.js#L55)... 

`record.debugJSON()` returns the `recordData`'s, `_data`, the last known data from the server inside `recordData._data`... so it returns the [reference](https://github.com/hjdivad/ember-m3/blob/971f3764b1aa7121e34ade4df82eaea8026d8b75/addon/record-data.js#L923) and [this method](https://github.com/hjdivad/ember-m3/blob/971f3764b1aa7121e34ade4df82eaea8026d8b75/addon/adapters/m3-debug-adapter.js#L35) is modifying the reference, which leads to this bug... this is _always_ happens, and leads to other kind of more "serious" bugs like changedAttributes stuff, maybe this also helps #769 ?

Screen Shot 2020-08-02 at 14 48 23

also since m3 lazily access/calculates stuff via unknownProperty , in the scenario of a corrupted/stringified `recordData._data`, accessing the props returns the stringified value. 

Screen Shot 2020-08-02 at 15 14 39

The easy solution _for now_ is just cloning, what do you think @SYU15 @igorT ?

### DIFF
--- a/addon/adapters/m3-debug-adapter.js
+++ b/addon/adapters/m3-debug-adapter.js
@@ -25,7 +25,7 @@ export default class M3DebugAdapter extends DataAdapter {
   }
 
   /**
-    Iterates through objects and if there is a nested object or array, stringifies the value
+    Clones the nestedJSON to avoid mutating references then iterates through objects and if there is a nested object or array, stringifies the value
     This is needed because Ember Inspector does not support data structures more than two levels deep
     @private
     @method _stringifyNestedValues
--- a/addon/adapters/m3-debug-adapter.js
+++ b/addon/adapters/m3-debug-adapter.js
@@ -33,6 +33,7 @@ export default class M3DebugAdapter extends DataAdapter {
     @return Object with stringified values as needed
   */
   _stringifyNestedValues(nestedJSON) {
+    nestedJSON = JSON.parse(JSON.stringify(nestedJSON));
     for (let key in nestedJSON) {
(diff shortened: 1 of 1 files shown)


addon/adapters/m3-debug-adapter.js
```diff
@@ -33,6 +33,7 @@ export default class M3DebugAdapter extends DataAdapter {
     @return Object with stringified values as needed
   */
   _stringifyNestedValues(nestedJSON) {
```
SYU15, line 35: Let's update this function's description since the behavior is probably counter-intuitive to an average reader. Also can you add a quick test? Can these steps be simulated by an integration test possibly (not sure tbh)? Thanks!

betocantu93: Sorry @SYU15 been a busy week, I just updated the comment, and I'm trying to create some integration tests, somehow I'm unable to replicate it on tests, gonna dig a little bit more I think, I'll keep you posted

hjdivad:
@betocantu93 @SYU15 thanks so much for digging in to this issue!  This looks good I'll be very happy to merge once @SYU15's comments are addressed.

Thanks again

PR #820, opened 2020-08-03
ci: Drop Node 13 and add Node 14

rwjblue:
Node 13 was EOL at the end of May this year. This doesn’t really change anything regarding our supported Node’s, but it ensures that CI is running on the platforms we support.

https://nodejs.org/en/about/releases/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


rwjblue:
@hjdivad - I personally think this is fine to land as a non-breaking change, but wanted to double check with you. Should we consider this a breaking change (if yes, I'll update our `package.json`'s `engines` field as well)?

hjdivad:
@rwjblue i agree

PR #836, opened 2020-08-20
Avoid checking project dependencies more than once per project.

rwjblue:
The original code was written to validate that the project has the required dependencies inside the `included` hook. This works perfectly fine and is _not_ overly slow (takes ~500ms in a **massive** application). This `included` hook is called once for every `ember-m3` addon instance in the addon graph.

Unfortunately, in some applications you may have a shared `data-layer` package that depends on `ember-m3`, and that `data-layer` package is depended on hundreds of other dependencies. In that scenario, the 500ms cost is multiplied by each `ember-m3` instance which results in a much larger overall cost (~ 200s).

The good news is that this is **very easy** for us to resolve! We are already checking the global project instance (which is the same across all of the `ember-m3` instances), so we just need to ensure we only do the validation once for each unique project.

That is exactly what this change does: it adds a simple `WeakSet` cache to ensure we only do this check one time per-project.

Note: this effectively reverts the changes in #834.


igorT:
@rwjblue when would the project be different?

rwjblue:
@igorT - When you build multiple projects within the same build. Which happens in a number of applications that I work on.

index.js
```diff
@@ -6,8 +6,14 @@ const getDebugMacros = require('./src/debug-macros').debugMacros;
...
-function assertValidEmberData(addon) {
-  let checker = VersionChecker.forProject(addon.project);
+const CHECKED_PROJECT = new WeakSet();
+
+function assertValidEmberData(project) {
+  if (CHECKED_PROJECT.has(project)) {
+    return;
+  }
+
+  let checker = VersionChecker.forProject(project);
```
igorT, line 16: Should the VersionChecker itself be caching this data per project?

PR #851, opened 2020-08-28
queryURL can resolve primitive values

### LINKED ISSUE #750: add support for queryURL returning primitives

hjdivad:
Previously queryURL would only be able to resolve

  1) Ember Data compatible JSON:API documents and
  2) null or undefined

[fix #750]


igorT:
Can we add tests that test the whole flow including the ember-data portion of this, rather than just the query cache part.

PR #884, opened 2020-09-24
Fix keeping embedded records in sync inside projections

igorT:
If you removed a record from an array of nested records, we did not update the baseRecordData, which thus kept the original nested childRecordData. If you then pushed a new object in, it would merge the new object with the previously removed data, leading to incorrect results.

It seems very likely that there are other issues with childRecordData projection syncing, that we should investigate.


igorT:
Described the more general problem in https://github.com/hjdivad/ember-m3/issues/885

PR #892, opened 2020-09-29
Fix keeping embedded records in sync inside projections

igorT:


igorT:
Backport of https://github.com/hjdivad/ember-m3/pull/884

PR #1005, opened 2021-02-08
Pass owner when creating M3DebugAdapter

### LINKED ISSUE #1001: Nightly Run Failure: 543249688
Nightly run failed on: 6 Feb 2021
https://github.com/hjdivad/ember-m3/actions/runs/543249688

pzuraq:
This should fix the recent failures on canary, which are due
to the M3DebugAdapter now eagerly looking up injections.

Fixes #1001

### DIFF
--- a/addon/adapters/interop-debug-adapter.js
+++ b/addon/adapters/interop-debug-adapter.js
@@ -1,4 +1,5 @@
 import M3DebugAdapter from './m3-debug-adapter';
+import { getOwner, setOwner } from '@ember/application';
 import { get, defineProperty } from '@ember/object';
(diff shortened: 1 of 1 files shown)


hjdivad:
is there a reasonable unit test we can add?

hjdivad (approved):
LGTM although if there's a reasonable unit test we can add that'd be nice, but if not the nightly canary tests are good enough.

rwjblue:
Since the issue is essentially due to a change in the underlying data adapter (it now happens to use `owner` where it didn't before) I don't think there is a test that isn't silly that we can write here... 🤔

PR #1022, opened 2021-03-06
only execute callback once per key for eachAttribute

betocantu93:
Hello, I just noticed something I consider a bug, but not sure if it's actually a bug, or intended.

When `MegamorphicModel.eachAttribute` is called, the callback is executed more than once per `key` if its on multiple "buckets" i.e. `this._attributes` and `this._data`

Screen Shot 2021-03-06 at 11 38 39

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -616,19 +616,22 @@ export default class M3RecordData {
       return this._baseRecordData.eachAttribute(callback, binding);
     }
 
+    let attrs = {};
     if (this.__attributes !== null) {
-      Object.keys(this._attributes).forEach(callback, binding);
+      Object.keys(this._attributes).forEach((key) => (attrs[key] = true));
     }
 
     if (this.__inFlightAttributes !== null) {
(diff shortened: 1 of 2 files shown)


addon/record-data.js
```diff
@@ -616,19 +616,22 @@ export default class M3RecordData {
...
     }
 
     if (this.__data !== null) {
       this._schema
         .computeAttributes(Object.keys(this._data), this.modelName)
-        .forEach(callback, binding);
+        .forEach((key) => (attrs[key] = true));
     }
+
+    Object.keys(attrs).forEach(callback, binding);
```
hjdivad, line 634: LGTM

Can you add a test for this? (e.g. local value for existing attr, trigger callback and assert the counts are correct

hjdivad (changes requested):

betocantu93:
@hjdivad sure thing!! added :)

hjdivad (approved):

hjdivad:
thanks @betocantu93!

hjdivad:
released as 3.0.7

PR #1031, opened 2021-03-12
fix: outer object replacement and nested property change bugfix

spham92:
## Bug
Top level object change is squashed by nested property change. We want the top level object change to remain intact.

## Recreate
The bug occurs when there's a nested object change such as:
```
let model = run(() => {
  return this.store.push({
    data: {
      id: 1,
      attributes: {
        reports: {
          highlyRated: {
            reportOne: {
              author: 'JK Rowling',
              reportDate: '2020-01-01',
            },
          },
        },
      },
    },
  });
});

model.set('reports', {
    highlyRated: {
      reportOne: {},
    },
  });
```

And another nested property is changed such as:
```
model.get('reports.highlyRated.reportOne');
reportOne.set('author', 'William Shakespeare');
```

## Root cause
Line 697 of `addon/record-data#changedAttributes` replaces the original change indicator for `reports` set in `_changedAttributes` on line 667

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -692,7 +692,8 @@ export default class M3RecordData {
           if (Object.keys(childChangedAttributes).length > 0) {
             if (
               this.getServerAttr(childKey) !== null &&
-              this.getServerAttr(childKey) !== undefined
+              this.getServerAttr(childKey) !== undefined &&
+              newData[childKey] === undefined // If object is not already staged for change
             ) {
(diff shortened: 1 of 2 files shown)


hjdivad (approved):
LGTM; will merge once github status is good to go again https://www.githubstatus.com/

hjdivad:
This has actually been merged; github is in a confused state

PR #1032, opened 2021-03-12
Adds support for native Proxy behind a canary feature flag

pzuraq:
Adds support for wrapping M3 models and arrays in a native Proxy,
allowing them to be accessed using Plain JS (tm) syntax. This is still
a work in progress as some functionality, such as `Array.isArray`, does
not work on the arrays, and additionally it is double wrapping
everything with Ember's own proxy system so it is very expensive
overall to enable perf-wise. As such, it is kept behind a feature flag
that can only be enabled in canary builds.

### DIFF
--- a/addon/base-record-array.js
+++ b/addon/base-record-array.js
@@ -14,9 +14,10 @@ import {
   deferPropertyChange,
   flushChanges,
 } from './utils/notify-changes';
-import { CUSTOM_MODEL_CLASS } from 'ember-m3/-infra/features';
+import { CUSTOM_MODEL_CLASS, PROXY_MODEL_CLASS } from 'ember-m3/-infra/features';
 import { recordDataToRecordMap, recordToRecordArrayMap } from './utils/caches';
 import { recordIdentifierFor } from '@ember-data/store';
+import { assert } from '@ember/debug';
 
 /**
  * BaseRecordArray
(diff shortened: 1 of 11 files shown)


addon/base-record-array.js
```diff
@@ -24,6 +24,43 @@ import { recordIdentifierFor } from '@ember-data/store';
...
+if (PROXY_MODEL_CLASS) {
+  const convertToInt = (prop) => {
+    if (typeof prop === 'symbol') return null;
+
+    const num = Number(prop);
+
+    if (isNaN(num)) return null;
+
+    return num % 1 === 0 ? num : null;
+  };
```
hjdivad, line 38: should this be an arrow function?  we don't really need the binding do we?

isn't `function(prop) { /* ... */ }` more appropriate?

addon/base-record-array.js
```diff
@@ -166,6 +212,15 @@ if (CUSTOM_MODEL_CLASS) {
...
     // public RecordArray API
+    static create(...args) {
+      let instance = super.create(...args);
+
+      if (PROXY_MODEL_CLASS) {
+        return new Proxy(instance, new BaseRecordArrayProxyHandler());
+      }
+
+      return instance;
+    }
```
hjdivad, line 223: I don't think we want this; this is an error condition.

`PROXY_MODEL_CLASS` requires `CUSTOM_MODEL_CLASS`

addon/base-record-array.js
```diff
@@ -299,6 +354,35 @@ if (CUSTOM_MODEL_CLASS) {
...
+
+    splice(idx, amt, ...values) {
+      return this.replace(idx, amt, values);
+    },
+
+    reverse() {
+      return this.reverseObjects();
+    },
+  });
+}
```
hjdivad, line 384: why on `BaseRecordArray` and not the proxy?

src/debug-macros.js
```diff
@@ -45,9 +45,15 @@ function getFlags(app, isDevelopingAddon) {
...
   features = app.project.require('@ember-data/private-build-infra/src/features')(isProd);
 
-  features = Object.assign({ CUSTOM_MODEL_CLASS: false }, features);
+  features = Object.assign(
+    {
+      CUSTOM_MODEL_CLASS: false,
+      PROXY_MODEL_CLASS: false,
+    },
+    features
+  );
```
hjdivad, line 54: this seems odd; how do we enable this at build time?  looks like we're pretending it's an Ember Data feature?

hjdivad (changes requested):

hjdivad (approved):

PR #1047, opened 2021-03-20
Refactor warning capturing and testing.

rwjblue:
Ember 3.27+ no longer compiles module imports down to globals, this means that when `import { warn } from '@ember/debug'` you **can no longer** stub `Ember.warn` directly and expect it to intercept the warning.

This commit introduces a new API that we can use to intercept warning invocations (leveraging `registerWarnHandler` from Ember itself), as well as ensuring that our own tests never "leak" unhandled warnings (which we didn't have a problem with, but seemed like a nice thing to add :smile:).

Fixes https://github.com/hjdivad/ember-m3/issues/1046

### DIFF
--- a/tests/helpers/warning-handler.js
+++ b/tests/helpers/warning-handler.js
@@ -1,0 +1,27 @@
+import { registerWarnHandler } from '@ember/debug';
+
+let errorOnInternalWarnings = true;
+export let capturedWarnings = null;
+
+export function captureWarnings() {
+  capturedWarnings = [];
+  errorOnInternalWarnings = false;
+}
+
+export function resetWarnings() {
+  capturedWarnings = null;
+  errorOnInternalWarnings = true;
+}
+
+registerWarnHandler(function (message, options, next) {
+  if (options.id.includes('ember-m3')) {
+    if (errorOnInternalWarnings) {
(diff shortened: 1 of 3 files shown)


tests/helpers/warning-handler.js
```diff
@@ -0,0 +1,22 @@
...
+    if (errorOnInternalWarnings) {
+      throw new Error(`Unhandled Warning: ${message}`);
+    }
+
+    // gather warnings for later assertion
+    capturedWarnings.push([message, options]);
+  } else {
+    next(message, options);
+  }
+});
```
hjdivad, line 22: seems weird that there's no inverse of `captureWarnings`?  should  we have a `reset`?

As it stands whether a warning in a test that never invokes `captureWarnings` throws an error on warning depends on test-ordering (i.e. is there a prior test that *did* invoke `captureWarnings`).

rwjblue: Good catch! Definitely a mistake.

hjdivad (approved):

rwjblue:
Confirmed this fixed the failures: https://github.com/hjdivad/ember-m3/actions/runs/679852171

PR #1099, opened 2021-05-12
Fix state notifications when CUSTOM_MODEL_CLASS is active

runspired:
without this check and early return we take property codepaths and then blow up in `assertNoChanges` because the keys are never handled.

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -276,7 +276,7 @@ export default class MegamorphicModel extends EmberObject {
       // just super and move on for state flags
       // this needs to match whatever we are notifying
       // in our subscription to the notificationManager
-      if (['isNew', 'isDeleted'].contains(key)) {
+      if (['isNew', 'isDeleted'].indexOf(key) !== -1) {
         super.notifyPropertyChange(key);
         return;
       }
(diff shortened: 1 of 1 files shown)


addon/model.js
```diff
@@ -272,6 +272,15 @@ export default class MegamorphicModel extends EmberObject {
...
 
   notifyPropertyChange(key) {
+    if (CUSTOM_MODEL_CLASS) {
+      // just super and move on for state flags
+      // this needs to match whatever we are notifying
+      // in our subscription to the notificationManager
+      if (['isNew', 'isDeleted'].indexOf(key) !== -1) {
+        super.notifyPropertyChange(key);
+        return;
+      }
```
hjdivad, line 282: Why are these the only state flags we need to exit early for? do we not have a similar issue with `record.reload()` and `isLoading` / `isLoaded`?

hjdivad, line 282: can we get a test for this please?

igorT: Currently ember-data notifies property change for `isNew` and `isDeleted`. Loading doesn't get notified, it is computed by us. We might want to expand this preemptively though

runspired: Specifically EmberData notifies state change, ember-M3 chooses to notify only these flags on state change. I could believe it needs to notify more things.

runspired: fwiw something like 258 tests failed without this, and will fail again if anything new needs to be added to the list.

PR #1100, opened 2021-05-12
Drop support for inactive node 10

hjdivad:


hjdivad:
`ember-qunit` -> `ember-test-loader` -> `workerpool` resulted in a version incompatible with node16.

PR #1122, opened 2021-06-01
Add Embroider testing setup

rwjblue:


rwjblue:
Embroider has an issue with our import of `@ember-data/record-data/-private` (likely because it can't figure out where that module actually is).

PR #1145, opened 2021-06-22
feat: make errors attribute configurable

spham92:
Adding the ability to configure whether to read from underlying data's `errors` attribute if it exists

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -641,7 +641,9 @@ export default class MegamorphicModel extends EmberObject {
   // Errors hash that will get update,
   // upon validation errors
   get errors() {
-    if (this._errors === null) {
+    if (this._schema.useUnderlyingErrorsValue(this._modelName)) {
+      return this.unknownProperty('errors');
+    } else if (this._errors === null) {
       this._errors = Errors.create();
     }
     return this._errors;
--- a/addon/services/m3-schema-manager.js
+++ b/addon/services/m3-schema-manager.js
@@ -57,6 +57,17 @@ export default class SchemaManager extends Service {
   }
 
   /**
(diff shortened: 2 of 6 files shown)


hjdivad:
@spham92 couple test failures with custom model classes it seems

addon/model.js
```diff
@@ -642,7 +642,16 @@ export default class MegamorphicModel extends EmberObject {
...
+      if (this._schema.useUnderlyingErrorsValue(this._modelName)) {
+        const errorsValue = this.unknownProperty('errors');
+        if (errorsValue) {
+          return errorsValue;
+        } else {
+          this._errors = Errors.create();
+        }
+      } else {
+        this._errors = Errors.create();
+      }
```
hjdivad, line 654: if `useUnderlyingErrorsValue() === true` why do we fall back to the `DS.Errors` object at all?
2. related, I think this check should come before the `null` check. It doesn't make a huge difference, but if the user is opting-out of this check it's weird to avoid the normal caching that `this.unknownProperty` does.

addon/services/m3-schema-manager.js
```diff
@@ -56,6 +56,17 @@ export default class SchemaManager extends Service {
...
+   * Calls the schema's useUnderlyingErrorsValue passing in modelName
+   *
+   * @param {string} modelName - Name of model to determine if `errors` property in the payload should be used
+   * @returns {boolean}
+   */
+  useUnderlyingErrorsValue(modelName) {
+    let schema = this.get('schema');
+    return schema.useUnderlyingErrorsValue(modelName);
+  }
+
```
hjdivad, line 69: This is great; can we add it to README please?

hjdivad (changes requested):

PR #1154, opened 2021-06-24
Remove husky

hjdivad:
It had broken since the upgrade from 4.x to 6.x and left some repos in a
bad state.  It provides no general value; contributors can set up their own
pre-commit hooks if they want, any other checks should be handled via
the CI test jobs (i.e. via yarn test).


rwjblue:
Seems like this accidentally added a `.husky/_/husky.sh` file to the repo (likely by uninstalling husky, it no longer adds that file to `.git/excludes` or something).

PR #1141, opened 2021-06-18
WIP - actions testing 

igorT:

PR #1157, opened 2021-06-25
Add CI workflow for running TracerBench

igorT:
Adapted from ember-data's performance CI. 
This will pass start working once https://github.com/hjdivad/ember-m3/pull/1156 is merged. In the meantime you can go to https://github.com/hjdivad/ember-m3/pull/1141 to see it in action


scripts/create-comment.js
```diff
@@ -0,0 +1,44 @@
+// Adapted from ember-data https://github.com/emberjs/data/blob/master/scripts/perf-tracking/create-comment.js
```
rwjblue, line 1: Please add `scripts/` to this list, then it'll fix the linting issues:

https://github.com/hjdivad/ember-m3/blob/1e8f75c397228b97cd014cd367ab5688ff392ba7/.eslintrc.base.js#L21-L35

igorT: Good call, done and fixed/linting now passes

scripts/post-comment.sh
```diff
@@ -0,0 +1,83 @@
...
+
+URI=https://api.github.com
+API_VERSION=v3
+API_HEADER="Accept: application/vnd.github.${API_VERSION}+json; application/vnd.github.antiope-preview+json"
+AUTH_HEADER="Authorization: token ${GITHUB_TOKEN}"
+
+FOUND_EXISTING=1
+update_comment_if_exists() {
+  # Get all the comments for the pull request.
+  body=$(curl -sSL -H "${AUTH_HEADER}" -H "${API_HEADER}" "${URI}/repos/${GITHUB_REPOSITORY}/issues/${NUMBER}/comments")
```
rwjblue, line 40: Future task: I'd like to see this script converted to using `gh` (which is setup and authorized) instead of `curl`.

.github/workflows/perf-check.yml
```diff
@@ -0,0 +1,61 @@
...
+                "markers": "start-loading,pushed-payload,end-loading"
+              }
+            }
+          fidelity: 60
+          upload-traces: true
+          upload-results: true
+      - name: Report TracerBench Results
+        if: failure() || success()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
rwjblue, line 55: Just to be clear, the usage of `${{ secrets.* }}` here means that this job will only run on PRs that are _from_ a branch on `hjdivad`'s repo (e.g. not from forks). Is that correct?

scripts/post-comment.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+# Copied from ember-data https://github.com/emberjs/data/blob/master/scripts/asset-size-tracking/src/post-comment.sh
```
rwjblue, line 3: Can you make this either a perma link, or link to some specfici tag? Over time this link may become invalid...

scripts/create-comment.js
rwjblue: Can you update this link to either a perma link or a link to a tag?

.github/workflows/perf-check.yml
igorT: There isn't a way to do the report step I believe without the token. Is there an alternative approach you would suggest?

rwjblue:
Ready for a rebase

.github/workflows/perf-check.yml
rwjblue: We can't do the comment step, but we definitely could run TracerBench itself and have it pass/fail the build.

That way the comment would be additive / nice to have, but we still have the same fundamental feature set either way.

igorT: Going through github docs, it seems that it can be set to be available but only in a read only mode, which will make the job run, but the comment to fail.  https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

Thats what seems to happen for ember data as well, where for outside PRs, the perf runs but the comment fails to post

https://github.com/emberjs/data/pull/7607/checks?check_run_id=2891069546#step:5:14

rwjblue: Ya, so we'll need @hjdivad to set that up (since it's a personal repo, only he is an admin).

> Thats what seems to happen for ember data as well, where for outside PRs, the perf runs but the comment fails to post

Overall, this is fine, but I want to ensure that we have a reasonable threshold set that will fail the perf-check CI job (make the status line red, to force an admin to look at the results).

scripts/post-comment.sh
igorT: :+1: thats a great thing for someone in data to do as well, I'll open an issue

scripts/post-comment.sh
igorT: done

.github/workflows/perf-check.yml
igorT: Yup, I am going to watch the next few PRs to see what the current noise looks like, and then consult with @hjdivad once he is back

scripts/create-comment.js
igorT: Done

PR #1167, opened 2021-06-30
Allow custom ObjectProxy instances to be wrapped around an M3Model instance 

igorT:
We discovered instances in the wild of users wrapping models in an ember object proxy and returning those from schema hooks. https://github.com/hjdivad/ember-m3/pull/789 broke some of those use cases, by accessing `_internalModel` in all of the array creation flows, thus triggering Ember's object proxy property access assertion. This is a temporary fix for that behavior, which goes away in the custom model class branch.

### DIFF
--- a/addon/resolve-attribute-util.js
+++ b/addon/resolve-attribute-util.js
@@ -2,6 +2,7 @@ import { dasherize } from '@ember/string';
 import { recordDataFor } from './-private';
 import { EmbeddedMegamorphicModel, EmbeddedSnapshot } from './model';
 import { A } from '@ember/array';
+import { get } from '@ember/object';
 import ManagedArray from './managed-array';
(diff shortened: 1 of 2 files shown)


spham92 (approved):
lgtm, nothing stands out

rwjblue (commented):
This seems actively bad/wrong for our users to do, I'd personally prefer to just make this assert if we spot something that isn't our model class here.

If it is indeed going to be removed in the feature flag branch, can you add a test that is ran in that feature flag that confirms an error is thrown.

rwjblue (approved):
This makes me very sad still 😭 😞 😢

igorT:
@rwjblue in this case it's fine to return any value in here, but if it's a m3/ember-data model we treat it differently

igorT:
> This makes me very sad still 😭 😞 😢

@rwjblue https://github.com/hjdivad/ember-m3/pull/1226 should make you happier

PR #1169, opened 2021-06-30
Custom Model Classes: Ensure isDirty for embedded records doesn't recursively loop

igorT:
With Custom Model Class, we were notifying property changes for `isDirty`  on embedded models, which normally depend on their parents `isDirty` state, causing a loop.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -32,8 +32,8 @@ module.exports = function () {
           name: 'ember-lts-n-1',
           npm: {
             devDependencies: {
-              'ember-source': '~3.20.0',
-              'ember-data': '~3.20.0',
+              'ember-source': '~3.12.0',
+              'ember-data': '~3.16.0',
               '@ember-data/store': null,
               '@ember-data/debug': null,
               '@ember-data/model': null,
@@ -48,8 +48,8 @@ module.exports = function () {
           name: 'ember-lts',
           npm: {
             devDependencies: {
-              'ember-source': '~3.24.0',
-              'ember-data': '~3.24.0',
+              'ember-source': '~3.16.0',
+              'ember-data': '~3.16.0',
               '@ember-data/store': null,
               '@ember-data/debug': null,
               '@ember-data/model': null,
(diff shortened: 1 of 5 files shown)


config/ember-try.js
```diff
@@ -48,8 +48,8 @@ module.exports = function () {
           name: 'ember-lts',
           npm: {
             devDependencies: {
-              'ember-source': '~3.16.0',
-              'ember-data': '~3.16.0',
+              'ember-source': '~3.24.0',
+              'ember-data': '~3.24.0',
```
rwjblue, line 52: Let's do these changes in a dedicated PR, so we can have a nice entry in the changelog.

tests/unit/model/state-test.js
```diff
@@ -1,8 +1,10 @@
 import { module, test, skip } from 'qunit';
-import { setupTest } from 'ember-qunit';
-
+import { setupTest, setupRenderingTest } from 'ember-qunit';
 import { run } from '@ember/runloop';
 import { isArray } from '@ember/array';
+import Component from '@ember/component';
+import hbs from 'htmlbars-inline-precompile';
```
rwjblue, line 6:
```suggestion
import { hbs } from 'ember-cli-htmlbars';
```

rwjblue (changes requested):

config/ember-try.js
igorT: done

spham92 (approved):
lgtm

package.json
```diff
@@ -64,6 +64,7 @@
     "broccoli-funnel": "^3.0.8",
     "ember-cli-babel": "^7.26.6",
     "ember-cli-version-checker": "^5.1.2",
+    "ember-compatibility-helpers": "^1.2.4",
```
rwjblue, line 67: devDep

rwjblue (approved):
One minor thing, but other than that LGTM

package.json
igorT: done

PR #1172, opened 2021-06-30
Upgrade ember-lts versions to actual latest and latest-1

igorT:


rwjblue (approved):
Note: this is not breaking, our support policy is pretty explicit (in fact we don't even **technically** support `lts-n-1`, but it is nice to know if we happen to break it accidentally).

PR #1179, opened 2021-07-01
Make perfomance testing app more robust to slow tests

igorT:


m3-perf-testing-app/tests/acceptance/materializing-test.js
```diff
@@ -21,7 +21,7 @@ module('Acceptance | materializing', function (hooks) {
     searches.forEach((search, i) => {
       // Sample the array for correctness, otherwise the test generates too
       // many assertions and take too long
-      if (i % 100 !== 0) {
+      if (i % 400 !== 0) {
```
rwjblue, line 24: The main thing that is important is that we have a test that ensures that the `/materializing` route actually functions (so that when we tracerbench it, the test is valid). Even validating **one** set of these searches probably satisifies that concern.

tldr; let's make this even larger, maybe:

```suggestion
      if (i % 1000 !== 0) {
```

PR #1183, opened 2021-07-02
Optimize schema types lookup

igorT:
We were looking up the schemaType twice needlessly.


rwjblue (commented):
Hmm, didn't this miss an instance of `resolveSingleValue` that should be updated to pass the `valueType`?

https://github.com/hjdivad/ember-m3/blob/115134850a5eedca340e1a501e99064e100a76fb/addon/resolve-attribute-util.js#L187

igorT:
@rwjblue great catch. Need to add test coverage to make this fail

PR #1184, opened 2021-07-02
Use  operator instead of property access for existence checks

igorT:


rwjblue (approved):
Discussed with @igorT out of band here to figure out why this change was needed, and the main issue here is that when we (eventually) move to using native property access these `internalModel.getRecord` checks may trigger either deprecations or assertions.

PR #1210, opened 2021-07-21
Fix plain arrays resolving and add tests

igorT:
Previously we did not have test coverage of resolving plain arrays that contain nested objects or references. This PR adds test coverage and fixes a bug where we would not detect nested models or references inside a plain array.


igorT:
CI currently failing due to an unrelated issue with ember-data canary and ember cached decorator  https://github.com/ember-polyfills/ember-cached-decorator-polyfill/issues/94 Fixing that in ember-data

hjdivad (approved):

PR #1214, opened 2021-07-22
Make materializing route in perf app configurable

igorT:
Perf app tests were timing out on CI, as running dev builds on CI containers was too slow, causing testem to time out while the records were being materialized. Now for testing the perf app, we materialize fewer records.


igorT:
Performance check is expected to fail as we changed the format of the url.

hjdivad (approved):

PR #1238, opened 2021-07-30
Add array native access

igorT:
Allows managed arrays to be used as native JS arrays. Removes `PROXY_MODEL_CLASS` feature flag, due to a different implementation for object dot access compared to the one that was feature flagged.


tests/unit/model/native-access-arrays-test.js
```diff
@@ -0,0 +1,182 @@
...
+
+      assert.equal(
+        chapters.join(' '),
+        '<EmbeddedMegamorphicModel:undefined> <EmbeddedMegamorphicModel:undefined>',
+        `Array.join works`
+      );
+
+      if ([].at) {
+        assert.equal(chapters.at(-1).get('name'), 'The Boy Who Lived', `Array.at is implemented`);
+      }
```
spham92, line 137: Does this mean this only runs sometimes?

igorT: Only if the browser already supports `Array.prototype.at`. Chrome added it in latest version, I had to update Chrome before it would work.

igorT: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#browser_compatibility

spham92 (approved):
lgtm

PR #1240, opened 2021-07-30
Fix for projected models resolving in projected arrays, when CUSTOM_MODEL_CLASS is on

igorT:
This was untested but worked by accident due to how `internalModel` cache worked. Added test coverage and a more robust solution in the Custom Model Classes case

### DIFF
--- a/addon/utils/resolve.js
+++ b/addon/utils/resolve.js
@@ -45,7 +45,7 @@ export function resolveReferencesWithRecords(store, references) {
         // If we have a cached recordData with the same id, but we have not seen a record with the same { type, id } pair
         // We could be a projection, in which case we want to push in a projected record with the new type
         let cachedRD = store._globalM3RecordDataCache[reference.id];
-        if (!record && cachedRD) {
+        if (cachedRD) {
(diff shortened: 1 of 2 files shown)


addon/utils/resolve.js
```diff
@@ -37,7 +37,28 @@ export function resolveReferencesWithRecords(store, references) {
...
-        return store.peekRecord(dasherize(reference.type), reference.id);
+        let normalizedType = dasherize(reference.type);
+        let record = store.peekRecord(normalizedType, reference.id);
+        if (record) {
+          return record;
+        }
+        // If we have a cached recordData with the same id, but we have not seen a record with the same { type, id } pair
+        // We could be a projection, in which case we want to push in a projected record with the new type
+        let cachedRD = store._globalM3RecordDataCache[reference.id];
+        if (!record && cachedRD) {
```
spham92, line 48: Do we still need a check for `record` here?

igorT: good catch

igorT: Fixed

spham92 (approved):

PR #1241, opened 2021-07-31
fix: skip this.errors.remove if useUnderlyingErrorsValue

spham92:
Skip the code flow for `_removeError ` if `this._schema.useUnderlyingErrorsValue(this._modelName)` returns true

### DIFF
--- a/tests/unit/model/errors-attribute-test.js
+++ b/tests/unit/model/errors-attribute-test.js
@@ -1,12 +1,9 @@
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import { run } from '@ember/runloop';
-import EmberObject, { get, set, computed } from '@ember/object';
+import { get } from '@ember/object';
 import DefaultSchema from 'ember-m3/services/m3-schema';
 import { Errors as ModelErrors } from '@ember-data/model/-private';
-import { Errors as StoreErrors } from '@ember-data/store/-private';
-
-const Errors = ModelErrors || StoreErrors;
 
 class TestSchemaFlagOn extends DefaultSchema {
   includesModel(modelName) {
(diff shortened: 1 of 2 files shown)


tests/unit/model/errors-attribute-test.js
```diff
@@ -89,6 +95,25 @@ module('unit/model/errors-attribute', function (hooks) {
...
+          id: 'urn:book:1',
+          type: 'com.example.bookstore.Book',
+          attributes: {
+            author: ['urn:author:1'],
+          },
+        },
+      });
+    });
+    const getErrorsSpy = this.sinon.spy(model, 'errors', ['get']);
+    model._removeError('author');
```
igorT, line 113: These kinds of tests couple us to current specific implementation. If we refactor `_removeError` to a different method, this test will fail, even though the behavior might still be correct. Can we write this test to assert user observable behavior?

igorT (changes requested):

tests/unit/model/errors-attribute-test.js
spham92: Ended changing to use `model.setUnknownProperty('title', 'Harry Potter');` which calls `_removeError`

igorT: It seems like you are still testing the internal implementation though. The user visible behavior here, is that if `useUnderlyingErrorsValue` is `true` then you can access `errors` after setting a property. If it's false, and you set a prop, the error gets removed, right?

igorT: Lets also create the errors using the public api and not by setting the property directly.  The `save errors getting updated via the store and removed upon setting a new value` has an example of how to get invalid errors. I've cleaned up that test as the setup wasn't quite correct here: https://github.com/hjdivad/ember-m3/pull/1251

igorT (approved):

PR #1242, opened 2021-07-31
Fix isDirty for inflight records and set _topModel to the proxy value

igorT:
In CUSTOM_MODEL_CLASS branches, we were not triggering the `isDirty` notification when the record had been saved.
Also, because we rely on object identity of `_topModel`, we need to account for potential proxy values that could be wrapping the Megamorphic model, so now we set '_topModel' after we instantiate the MegamorphicModel instance.

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -141,6 +141,7 @@ if (CUSTOM_MODEL_CLASS) {
 export default class MegamorphicModel extends EmberObject {
   static create(...args) {
     let instance = super.create(...args);
+    let value = instance;
     if (CUSTOM_MODEL_CLASS) {
       let useNative = instance._schema.useNativeProperties(instance._modelName);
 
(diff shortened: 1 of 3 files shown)


spham92 (approved):

addon/model.js
```diff
@@ -149,22 +150,23 @@ export default class MegamorphicModel extends EmberObject {
...
+          value = proxy;
         }
       }
     }
-
-    return instance;
+    if (!value._topModel) {
+      value._topModel = value;
+    }
+    value._flushInitProperties();
```
hjdivad, line 168: calling `_flushInitProperties` in `create` is sus.

Are we even still batching the initial properties?

I think it's possible this isn't needed with the versions of ember in support.

hjdivad: hmm well I guess I'm wrong, we still need to batch, it's just that we now have a stable call order for the sets of versions we support.

igorT: it used to be in the constructor, but we want to now call it after we have set the `_topModel` to avoid double computing values.

PR #1249, opened 2021-08-02
Fix isSaving for embedded records when CUSTOM MODEL CLASS is on

igorT:
We can't use the request service to lookup the embedded records. They inherit the saving state of their parent.


igorT:
This needs the `_topModel` fix from https://github.com/hjdivad/ember-m3/pull/1242 to pass CI

spham92 (approved):
lgtm

PR #1291, opened 2021-08-23
Add support for tracked properties with proxied managed array

igorT:
Because ManagedArray is now behind a proxy, it does not pass the internal ember `isEmberArray` check that uses a WeakSet https://github.com/emberjs/ember.js/blob/21bd70c773dcc4bfe4883d7943e8a68d203b5bad/packages/%40ember/-internals/utils/lib/ember-array.ts, so we have to access '[]' when we access ManagedArrays

### DIFF
--- a/addon/base-record-array.js
+++ b/addon/base-record-array.js
@@ -26,6 +26,8 @@ import { recordIdentifierFor } from '@ember-data/store';
 let BaseRecordArray;
 let baseRecordArrayProxyHandler;
 
+const MANAGED_ARRAYS = new WeakSet();
+
 if (CUSTOM_MODEL_CLASS) {
   const convertToInt = (prop) => {
     if (typeof prop === 'symbol') return null;
@@ -77,7 +79,9 @@ if (CUSTOM_MODEL_CLASS) {
     static create(...args) {
       let instance = super.create(...args);
 
-      return new Proxy(instance, baseRecordArrayProxyHandler);
+      let proxy = new Proxy(instance, baseRecordArrayProxyHandler);
+      MANAGED_ARRAYS.add(proxy);
+      return proxy;
     }
 
     init() {
@@ -108,6 +112,7 @@ if (CUSTOM_MODEL_CLASS) {
     }
 
     objectAt(idx) {
+      get(this, '[]');
       this._resolve();
       // TODO make this lazy again
       let record = this._objects[idx];
@@ -206,6 +211,7 @@ if (CUSTOM_MODEL_CLASS) {
     }
 
     get length() {
+      get(this, '[]');
       return this._resolved ? this._objects.length : this._references.length;
     }
   };
@@ -375,8 +381,8 @@ if (CUSTOM_MODEL_CLASS) {
       return this.popObjects(values);
     },
 
-    shift(...values) {
-      return this.shiftObjects(values);
+    shift() {
+      return this.shiftObject();
     },
 
     unshift(...values) {
@@ -438,3 +444,4 @@ export function associateRecordWithRecordArray(record, recordArray) {
 }
 
 export default BaseRecordArray;
+export { MANAGED_ARRAYS };
--- a/addon/base-record-array.js
+++ b/addon/base-record-array.js
@@ -112,6 +112,8 @@ if (CUSTOM_MODEL_CLASS) {
     }
 
     objectAt(idx) {
+      // Need to get '[]' in order to entangle tracked properties, as ember won't treat us as an ember array when we are a proxy
+      // See https://github.com/emberjs/ember.js/issues/19139#issuecomment-694487616 for context
       get(this, '[]');
       this._resolve();
       // TODO make this lazy again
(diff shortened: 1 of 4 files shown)


addon/base-record-array.js
```diff
@@ -108,6 +112,7 @@ if (CUSTOM_MODEL_CLASS) {
     }
 
     objectAt(idx) {
+      get(this, '[]');
```
hjdivad, line 115: this needs an explanatory comment

addon/base-record-array.js
```diff
@@ -206,6 +211,7 @@ if (CUSTOM_MODEL_CLASS) {
     }
 
     get length() {
+      get(this, '[]');
```
hjdivad, line 214: explanatory comment please

addon/model.js
```diff
@@ -145,6 +146,16 @@ if (CUSTOM_MODEL_CLASS) {
...
 
+// If a value returned from unknownProperty is a ManagedArray we need to access '[]'
+// to keep parity with how ember treats arrays
+// see https://github.com/emberjs/ember.js/blob/3ce13cea235cde8a87d89473533c453523412764/packages/%40ember/-internals/metal/lib/property_get.ts#L136
+function accessBracketsIfArray(maybeArray) {
+  if (MANAGED_ARRAYS.has(maybeArray)) {
+    get(maybeArray, '[]');
+  }
+  return maybeArray;
+}
```
hjdivad, line 157: Let's rename this to something like `readCachedValue(key)`

As it stands the naming is odd. You get usage like

```js
  if (key in this._cache) {
      return accessBracketsIfArray(this._cache[key]);
    }
```

So here the call site cares about the return value, but is calling a function named for its side-effects. This is a kind of abstraction leakage -- contrast to

```js
  if (key in this._cache) {
      return readCachedValue(this._cache, key);
    }
```

now the call site is asking for what it needs in order to return a value (the cached value) and the side effects necessary to entangle the array via ember's semantics is encapsulated in the implementation of `readCachedValue`.

tests/unit/model/native-access-arrays-test.js
```diff
@@ -99,17 +99,6 @@ if (CUSTOM_MODEL_CLASS) {
...
-      let shiftedChapter = shiftedChapters.shift();
-
-      assert.equal(shiftedChapters[0].get('name'), 'The Boy Who Lived', `Array.shift works`);
-      assert.equal(shiftedChapters.length, 1, `Array.shift works`);
-      assert.equal(shiftedChapter.get('name'), 'The Vanishing Glass', 'Array.shift works');
-
-      shiftedChapters.unshift(shiftedChapter);
-      assert.equal(shiftedChapters[0].get('name'), 'The Vanishing Glass', `Array.unshift works`);
-      assert.equal(shiftedChapters.length, 2, `Array.unshift works`);
-
```
hjdivad, line 112: why are these removed?

tests/unit/model/native-access-arrays-test.js
```diff
@@ -99,17 +99,6 @@ if (CUSTOM_MODEL_CLASS) {
       assert.equal(slicedChapters[0].get('name'), 'The Boy Who Lived', `Array.slice still works`);
```
hjdivad, line 99: why do we use `get` in these tests?

hjdivad (changes requested):

tests/unit/model/native-access-arrays-test.js
igorT: To not depend on `useNativeProperties` being true

tests/unit/model/native-access-arrays-test.js
igorT: These have been moved to https://github.com/hjdivad/ember-m3/pull/1291/files/ea086c142eeb5656dcd4f8135c5038fa4ffaeae7#diff-ca4b95cfbe6d1d2a7c6d1b9d8d44b4593ee6b0470174d1e49b004bfd14bb65b1R171 to simplify the tests

tests/unit/model/native-access-arrays-test.js
hjdivad: do we test the common case of using both of these?

addon/base-record-array.js
igorT: done

addon/base-record-array.js
igorT: done

addon/model.js
igorT: done

tests/unit/model/native-access-arrays-test.js
igorT: The tests added here use them together.

addon/model.js
hjdivad: what's the reason for preferring `readCachedValue(value)` over the API I suggested above?

tests/unit/model/native-access-arrays-test.js
hjdivad: it's just `'mutating reference arrays cause re-renders'` isn't it?  the other two use `get`.

I guess it's probably okay, I don't have any particular reason to think these two features don't work well together, I just find it a bit odd that we don't have many tests for the most common scenario

igorT: Happy to add more separately.

addon/model.js
igorT: Discussed on slack, updated the method now so we don't have a separate function call

hjdivad (approved):
LGTM please rebase with a clean  commit history before merging

PR #1295, opened 2021-08-27
Fix isDeleted rollbackAttributes for Custom Model Class

igorT:
`isDeleted` state used to be handled by ember-data internal-model, with `3.28` we now need to handle it ourselves

### DIFF
--- a/tests/unit/model/state-test.js
+++ b/tests/unit/model/state-test.js
@@ -138,7 +138,29 @@ for (let testRun = 0; testRun < 2; testRun++) {
         assert.equal(record.get('isLoaded'), true, 'record is loaded');
       });
       skip('isSaving', function () {});
-      skip('isDeleted', function () {});
+      test('isDeleted', function (assert) {
+        let data = {
+          data: {
+            id: 1,
+            type: 'com.example.bookstore.Book',
+            attributes: {
+              title: 'The Storm Before the Storm',
+              author: 'Mike Duncan',
+            },
+          },
+        };
+        let record = this.store.push(data);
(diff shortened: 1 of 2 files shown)


tests/unit/model/state-test.js
```diff
@@ -138,7 +138,29 @@ for (let testRun = 0; testRun < 2; testRun++) {
...
+        let record = this.store.push(data);
+        assert.equal(record.get('isDeleted'), false, 'record starts off not deleted');
+        record.deleteRecord();
+        assert.equal(record.get('isDeleted'), true, 'record is now deleted');
+        record.rollbackAttributes();
+        assert.equal(
+          record.get('isDeleted'),
+          false,
+          'after rollbackAttributes record is no longer deleted'
+        );
```
hjdivad, line 161: should we not write new tests with direct property access (i.e. what we expect our users to do) and not `get`?

hjdivad (commented):
LGTM but I'd prefer we write new tests to match what we expect users to do (i.e. using native-compatible access, getters & arrays)

tests/unit/model/state-test.js
igorT: We still want to cover the older ember-data versions as well. I can open a PR that switches between `.get` and prop access for these tests based on version

igorT: https://github.com/hjdivad/ember-m3/pull/1296

hjdivad: @igorT makes sense; presumably something similar to https://github.com/hjdivad/ember-m3/blob/c384658c2cae1320c306e9152a955b5750e08fab/tests/helpers/schema-versions.js#L41-L78

PR #1298, opened 2021-08-31
Do not needlessly access `isDirty`

igorT:
We were entangling `isDirty` by doing a `get(this, 'isDirty');` even when no one had read our dirty state. In some edge cases, this could result in rerendering errors, and was also not great for performance.

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -643,12 +643,21 @@ export default class MegamorphicModel extends EmberObject {
     schemaInterface._suppressNotifications = priorSuppressNotifications;
 
     const hasDirtyAttr = recordData.hasChangedAttributes();
-    const isDirty = get(this, 'isDirty');
+    if (!CUSTOM_MODEL_CLASS) {
+      const isDirty = get(this, 'isDirty');
 
-    if (hasDirtyAttr && !isDirty) {
-      this._updateCurrentState(!CUSTOM_MODEL_CLASS && updatedUncommitted);
-    } else if (!hasDirtyAttr && isDirty) {
-      this._updateCurrentState(!CUSTOM_MODEL_CLASS && loadedSaved);
+      if (hasDirtyAttr && !isDirty) {
+        this._updateCurrentState(updatedUncommitted);
+      } else if (!hasDirtyAttr && isDirty) {
+        this._updateCurrentState(loadedSaved);
+      }
+    } else {
+      const cachedIsDirty = this._topModel._isDirty;
+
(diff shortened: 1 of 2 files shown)


tests/unit/model/state-test.js
```diff
@@ -330,5 +331,34 @@ if (gte('3.24.0')) {
...
+          class ShowDirtyness extends Component {
+            layout = hbs`
+          {{#if this.myBook.isDirty}}
+            Book is dirty
+          {{/if}}
+      `;
+
+            get myBook() {
+              // Passing `{ someKey: undefined }` will trigger a property change on the newly created model, but will not dirty the properties
+              // If we are not careful we could create a rerender cycle by notifying `isDiry` change on a record in the middle of instantiation
```
spham92, line 348:
```suggestion
              // If we are not careful we could create a rerender cycle by notifying `isDirty` change on a record in the middle of instantiation
```

tests/unit/model/state-test.js
```diff
@@ -330,5 +331,34 @@ if (gte('3.24.0')) {
...
+          }
+        );
+
+        this.set('store', this.store);
+
+        await render(hbs`
+        {{show-dirtyness store=this.store}}
+      `);
+
+        assert.equal(this.element.innerText, 'Book is dirty', 'Book renders as dirty');
```
spham92, line 360: Do we expect this to never hit if there's a re-render cycle?

igorT: Yup, Ember would error out

spham92 (approved):

addon/model.js
```diff
@@ -643,12 +643,21 @@ export default class MegamorphicModel extends EmberObject {
...
+      } else if (!hasDirtyAttr && isDirty) {
+        this._updateCurrentState(loadedSaved);
+      }
+    } else {
+      const cachedIsDirty = this._topModel._isDirty;
+
+      // `isDirty` cached value does not match our current one -> we need to update our state
+      if (hasDirtyAttr !== cachedIsDirty) {
+        this._updateCurrentState();
+      }
```
hjdivad, line 660: Can we swap the order here (i.e. change the conditional to if `CUSTOM_MODEL_CLASS`) so it matches the order where we set the cached value.

Having the then/else branches switched makes this harder to read.

hjdivad (dismissed):

addon/model.js
igorT: Done

PR #1299, opened 2021-08-31
Switch m3 array proxies to proxy [] insted of BaseRecordArray

igorT:
The previous approach would result in apps breaking due to subtle differences in how ember entangles `[]` on template access vs us doing it in `objectAt`, `length` and `unknownProperty`.

This will be a breaking change for apps relying on `Array.isArray` returning false for m3 arrays. Using `addon/utils/is-m3-array.js` should enable those apps to migrate

### DIFF
--- a/addon/base-record-array.js
+++ b/addon/base-record-array.js
@@ -26,8 +26,6 @@ import { recordIdentifierFor } from '@ember-data/store';
 let BaseRecordArray;
 let baseRecordArrayProxyHandler;
 
-const MANAGED_ARRAYS = new WeakSet();
-
 if (CUSTOM_MODEL_CLASS) {
   const convertToInt = (prop) => {
     if (typeof prop === 'symbol') return null;
@@ -40,14 +38,17 @@ if (CUSTOM_MODEL_CLASS) {
   };
 
   const BaseRecordArrayProxyHandler = class {
+    getPrototypeOf(target) {
+      return Object.getPrototypeOf(target._instance);
+    }
     get(target, key, receiver) {
       let index = convertToInt(key);
 
       if (index !== null) {
-        return receiver.objectAt(key);
+        return target._instance.objectAt(key);
       }
 
-      return Reflect.get(target, key, receiver);
+      return Reflect.get(target._instance, key, receiver);
     }
 
     set(target, key, value, receiver) {
@@ -56,7 +57,7 @@ if (CUSTOM_MODEL_CLASS) {
       if (index !== null) {
         receiver.replace(index, 1, [value]);
       } else {
-        Reflect.set(target, key, value, receiver);
+        Reflect.set(target._instance, key, value);
       }
 
       return true;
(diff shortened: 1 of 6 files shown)


tests/integration/managed-array-tracked-test.js
```diff
@@ -134,6 +138,37 @@ if (CUSTOM_MODEL_CLASS) {
       );
     });
 
+    // We have run into scenarios like these with users stashing M3 Arrays on POJOs in services, and then accessing them and
+    // modifying during a render
+    test('Can modify a managed array after creation without triggering rerendering assertions', async function (assert) {
```
spham92, line 143: I'm assuming this test would fail when the ember infra throws an error ?

igorT: Yup, Ember would error out

spham92 (approved):
lgtm

addon/base-record-array.js
```diff
@@ -79,9 +80,9 @@ if (CUSTOM_MODEL_CLASS) {
     static create(...args) {
       let instance = super.create(...args);
 
-      let proxy = new Proxy(instance, baseRecordArrayProxyHandler);
-      MANAGED_ARRAYS.add(proxy);
-      return proxy;
+      let arr = [];
+      arr._instance = instance;
```
hjdivad, line 84: can we rename `__instance` to be less cryptic? it's the record array right?  wouldn't `__recordArray` be more clear?

addon/base-record-array.js
```diff
@@ -199,6 +197,23 @@ if (CUSTOM_MODEL_CLASS) {
       });
     }
 
+    // Need to subclass `removeAt`, `pushObject`, and `insertAt` because the default implementations by
```
hjdivad, line 200:
```suggestion
    // Need to override `removeAt`, `pushObject`, and `insertAt` because the default implementations by
```

addon/base-record-array.js
```diff
@@ -199,6 +197,23 @@ if (CUSTOM_MODEL_CLASS) {
...
+      return this;
+    }
+
+    pushObject(obj) {
+      return this.insertAt(this.length, obj);
+    }
+
+    insertAt(idx, object) {
+      return this.replace(idx, 0, [object]);
+    }
```
hjdivad, line 215: shouldn't these all be deprecated as well? (i.e. non-native methods)?

igorT: Given how layered these methods are, it seems better to deprecate once in `replace` and `objectAt` rather than try to hit each method individually. I'll do that separately from this PR

addon/base-record-array.js
igorT: Done

addon/base-record-array.js
igorT: Does it make sense to deprecate the non-native methods? If you were using a JS array, you would be using the ember array methods for mutating it if you wanted the changes to be observable. If we deprecate them, it becomes harder to write functions that work on both JS arrays and M3 Arrays.

PR #1300, opened 2021-08-31
Do not emit array change events when app is being destroyed

igorT:
Speeds up test runs, and also makes app tests more robust against bad teardown states.

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -587,6 +587,10 @@ export default class M3RecordData {
 
   removeFromRecordArrays() {
     if (CUSTOM_MODEL_CLASS) {
+      // No need to remove us from record arrays if the store is beeing destroyed
+      if (this.globalM3CacheRD._storeIsDestroying) {
+        return;
+      }
       this._recordArrays.forEach((recordArray) => {
         recordArray._removeRecordData(this);
       });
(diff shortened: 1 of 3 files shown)


spham92 (approved):

addon/record-data.js
```diff
@@ -581,6 +581,10 @@ export default class M3RecordData {
 
   removeFromRecordArrays() {
     if (CUSTOM_MODEL_CLASS) {
+      // No need to remove us from record arrays if the store is beeing destroyed
+      if (this.globalM3CacheRD._storeIsDestroying) {
+        return;
+      }
```
hjdivad, line 587: This seems like an odd place to stash a store destroying flag. Does the store wrapper not know whether the store is being destroyed?

igorT: No, there is no public API for it, so I made one we create. Any ideas for a better place to put it?

hjdivad: putting the flag on the record data as a workaround seems fine to me as long as we agree that this is a deficiency in the public API that should be sorted out

igorT: Moved to using schema's `isDestroying` flag

hjdivad (approved):

PR #1314, opened 2021-09-01
Do not use an EmberArray for m3 array internal state

igorT:
We don't need to have `_objects` be tracked because we manage notifications ourselves. Having `_objects` be an `Ember.A()` just ends up triggering change notifications that don't go anywhere.


hjdivad (changes requested):
LGTM but CI red

igorT:
@hjdivad yup, CI caught an IE11 fail, looking into it

PR #1315, opened 2021-09-01
Add IE testing and guard proxies when they are not present

igorT:
We were not testing against IE11, and a few failures had snuck in. We now only create m3 array proxies when we are able to instantiate a Proxy class, similarly to how Ember does it https://github.com/emberjs/ember.js/blob/v3.28.0/packages/@ember/-internals/utils/lib/proxy-utils.ts

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "ember-compatibility-helpers": "^1.2.5",
     "ember-disable-prototype-extensions": "^1.1.2",
     "ember-export-application-global": "^2.0.1",
+    "ember-fetch": "^8.1.1",
     "ember-inflector": "^4.0.2",
     "ember-load-initializers": "^2.1.2",
     "ember-lodash": "^4.17.5",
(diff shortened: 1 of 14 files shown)


tests/dummy/app/adapters/-json-api.js
```diff
@@ -1,4 +1,5 @@
 import EmberObject from '@ember/object';
+import fetch from 'fetch';
```
hjdivad, line 2: what does this have to do with IE testing?

package.json
```diff
@@ -88,6 +88,7 @@
     "ember-compatibility-helpers": "^1.2.5",
     "ember-disable-prototype-extensions": "^1.1.2",
     "ember-export-application-global": "^2.0.1",
+    "ember-fetch": "^8.1.1",
```
hjdivad, line 91: what does this have to do with IE testing?

tests/dummy/app/adapters/-json-api.js
spham92: I think we need it so it polyfills for IE11?

https://caniuse.com/fetch

package.json
igorT: Need it for https://github.com/hjdivad/ember-m3/blob/master/tests/dummy/app/adapters/-json-api.js#L5

PR #1301, opened 2021-08-31
Sync projections and base model deleted state with CUSTOM_MODEL_CLASS

igorT:
Projections and base models now have a synced up deleted state.

PR #1317, opened 2021-09-02
Unify `isDeleted`, `isNew` and `isDeletionCommitted` flag handling on Record Data

igorT:
Building upon #1301, we now always keep record state flags on the `baseRecordData` and notify the record and projections when they change, having a unidirectional data flow. This is more robust, and fixes bugs where projections would get their state out of sync.

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -177,10 +177,13 @@ export default class M3RecordData {
       if (!baseRecordData && !parentRecordData && id) {
         this.globalM3CacheRD[this.id] = this;
       }
-      this._isNew = false;
-      this._isDeleted = false;
+      // We keep the state of the lifecycle flags on the base RD, and projections read from there
+      if (!baseRecordData) {
+        this._isNew = false;
+        this._isDeleted = false;
+        this._isDeletionCommitted = false;
+      }
       this._isLoaded = false;
-      this._isDeletionCommited = false;
     } else {
       this._embeddedInternalModel = null;
     }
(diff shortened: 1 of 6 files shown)


addon/record-data.js
```diff
@@ -497,7 +511,33 @@ export default class M3RecordData {
...
+    if (this.isDeletionCommitted()) {
+      this.removeFromRecordArrays();
+    }
+    if (record) {
+      record._updateCurrentState();
+    }
+    if (this._projections) {
+      for (let i = 1; i < this._projections.length; i++) {
+        this._projections[i]._notifyStateChange();
+      }
```
hjdivad, line 539: why do we skip `projections[0]`? that's the base (unprojected) case?

igorT: > that's the base (unprojected) case? -> Yup, it's a pattern we use in a few places I think

We always notify on the base case itself, so 
```
    let record = recordDataToRecordMap.get(this);
    if (this.isDeletionCommitted()) {
      this.removeFromRecordArrays();
    }
    if (record) {
      record._updateCurrentState();
    }
```
deals with the base and 
```
      for (let i = 1; i < this._projections.length; i++) {
        this._projections[i]._notifyStateChange();
      }
      ```
      with the projections

hjdivad: yeah it's unfortunate that we have to remember it at every loop of `this._projections` but oh well :shrug:

hjdivad (approved):

PR #1337, opened 2021-09-08
Add support for Node v16

igorT:
Node v16 is It is moving from current to active on October 26, we should allow it so people can test their apps with the upcoming node LTS


spham92 (approved):
lgtm

hjdivad (approved):

PR #1376, opened 2021-09-28
Fix property access in production and turn on prod testing

### LINKED ISSUE #1371: Native property access doesn’t work in production
When running m3 in a production build using native properties doesn’t work, as `megamorphicModelProxyHandler` isn’t defined, and the proxy constructor throws an error. It’s only defined inside an `if (DEBUG)` statement, making it work fine in development builds. 

Would it be sufficient to move the declaration outside of that if statement, or is there a reason it’s inside it?

Code ref: https://github.com/hjdivad/ember-m3/blob/master/addon/model.js#L115

igorT:
Fixes #1371

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -91,7 +91,7 @@ if (CUSTOM_MODEL_CLASS) {
       // have to use it. This is safe, because we already checked that the
       // property is not `in` the instance, so it will definitely call
       // `unknownProperty` and will not re-enter.
-      return receiver.get(key);
+      return target.get(key);
     }
 
     set(target, key, value, receiver) {
@@ -111,9 +111,8 @@ if (CUSTOM_MODEL_CLASS) {
     }
   };
 
+  megamorphicModelProxyHandler = new MegamorphicModelProxyHandler();
   if (DEBUG) {
-    megamorphicModelProxyHandler = new MegamorphicModelProxyHandler();
-
     const MegamorphicNativeDeprecationProxyHandler = class {
       // Need to implement the getter for the Ember Proxy assertions to work
       get(target, key) {
--- a/addon/model.js
+++ b/addon/model.js
@@ -91,6 +91,15 @@ if (CUSTOM_MODEL_CLASS) {
       // have to use it. This is safe, because we already checked that the
       // property is not `in` the instance, so it will definitely call
       // `unknownProperty` and will not re-enter.
+
+      // We have to use `target.get` instead of `receiver.get` because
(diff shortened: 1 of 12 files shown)


addon/model.js
```diff
@@ -91,7 +91,7 @@ if (CUSTOM_MODEL_CLASS) {
       // have to use it. This is safe, because we already checked that the
       // property is not `in` the instance, so it will definitely call
       // `unknownProperty` and will not re-enter.
-      return receiver.get(key);
+      return target.get(key);
```
hjdivad, line 94: 1. comment is now out of date
2. there's some trickiness here, the comment should explain
  i. why we're using `target` instead of `receiver`, since it's non-obvious without knowing the `Ember.get` implementation
  ii. what trade-off is (i.e. that `this` will not be the proxy in `unknownProperty` which is something we only need to be aware of within `ember-m3` but not something that will impact users)

tests/unit/model/native-access-test.js
```diff
@@ -122,124 +123,135 @@ if (CUSTOM_MODEL_CLASS && HAS_NATIVE_PROXY) {
...
+        let book = this.store.peekRecord('com.example.bookstore.Book', 'urn:li:book:1');
 
-      assert.equal(
-        book.title,
-        'How to Win Friends and Influence People',
-        'accessed field via native property get'
-      );
+        assert.expectDeprecation(() => {
+          book.name = 'Temp title';
+        }, `You set the property 'name' on a 'com.example.bookstore.book' with id 'urn:li:book:1'`);
```
hjdivad, line 173: why don't we `if(DEBUG)` this assertion rather than the whole test?  the bulk of  the test still seems relevant in prod mode or am i missing something?

hjdivad (commented):
LGTM, only real complaint is let's update the comment in the proxy as the code is a little tricky and it depends on particular details of `Ember.get`

tests/unit/model/native-access-test.js
igorT: Yeah good point, we would still want it to work in prod and I'm not sure we have a different test for it. I'll rename the test as well, I added the guard because the name suggested it was asserting the deprecation

addon/model.js
igorT: Fixed

tests/unit/model/native-access-test.js
igorT: Done

hjdivad (approved):
LGTM please combine the commits before merging so we have clean commit history

PR #1403, opened 2021-10-12
Fix for tracked properties updating when using m3 native properties

igorT:
The fix in https://github.com/hjdivad/ember-m3/pull/1376 was not correct. In order to properly entangle we need to be doing `receiver.get` but to protect against re-entrance in production, we need to manually keep track of the key being accessed.

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -115,20 +115,20 @@ if (CUSTOM_MODEL_CLASS) {
       // to trigger `unknownProperty`, as otherwise we will end up in an infinite loop
       // of repeatadly calling `receiver.get`
 
-     if (this.getting === key) {
-       return undefined;
-     }
-     
-     let shouldReset = this.getting === '';
-     this.getting = key;
-     
-     try {
-       return get(receiver, key);
-     } finally {
-       if (shouldReset) {
-         this.getting = '';
-       }
-     }
+      if (this.getting === key) {
+        return undefined;
+      }
+
+      let shouldReset = this.getting === '';
+      this.getting = key;
+
+      try {
+        return get(receiver, key);
+      } finally {
+        if (shouldReset) {
+          this.getting = '';
+        }
+      }
     }
 
     set(target, key, value, receiver) {
--- a/addon/model.js
+++ b/addon/model.js
@@ -115,14 +115,20 @@ if (CUSTOM_MODEL_CLASS) {
       // to trigger `unknownProperty`, as otherwise we will end up in an infinite loop
       // of repeatadly calling `receiver.get`
 
-      if (!DEBUG) {
-        if (this.getting === key) {
-          this.getting = undefined;
-          return undefined;
-        }
-        this.getting = key;
-      }
-      return get(receiver, key);
+     if (this.getting === key) {
+       return undefined;
+     }
+     
+     let shouldReset = this.getting === '';
+     this.getting = key;
+     
+     try {
+       return get(receiver, key);
+     } finally {
+       if (shouldReset) {
+         this.getting = '';
+       }
+     }
     }
 
     set(target, key, value, receiver) {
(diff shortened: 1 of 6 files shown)


addon/model.js
```diff
@@ -80,6 +80,10 @@ let megamorphicModelProxyHandler, megamorphicNativeDeprecationHandler;
 
 if (CUSTOM_MODEL_CLASS) {
   const MegamorphicModelProxyHandler = class {
+    constructor() {
+      this.getting = '';
+    }
```
rwjblue, line 85: Let's use class fields instead (easier to reason about):

```suggestion
    this.getting = '';
```

rwjblue (commented):
Given the explanation in the comments, I think there is probably a more straight forward solution for us.

Do we think (with the new tests added) that all of the cases that we care about are handled?

igorT:
> Do we think (with the new tests added) that all of the cases that we care about are handled
I am not aware of any, do any come to mind?

addon/model.js
```diff
@@ -88,19 +90,39 @@ if (CUSTOM_MODEL_CLASS) {
...
+      // of repeatadly calling `receiver.get`
+
+      if (!DEBUG) {
+        if (this.getting === key) {
+          this.getting = undefined;
+          return undefined;
+        }
+        this.getting = key;
+      }
+      return get(receiver, key);
```
rwjblue, line 125:
```suggestion
     this.getting = key;
     try {
       return get(receiver, key);
     } finally {
       this.getting = undefined;
     }
```

tests/unit/model/native-access/tracked-properties-test.js
```diff
@@ -0,0 +1,140 @@
...
+        renderedTitle,
+        'Title: How to Win Friends and Influence People',
+        'component is rendered correctly'
+      );
+      book.title = 'New title';
+      await settled();
+
+      renderedTitle = this.element.querySelector('.title').innerText;
+      assert.equal(renderedTitle, 'Title: New title', 'component is rendered correctly');
+    });
```
rwjblue, line 76: Can you add this exact test, but with the component:

```js
    test('Component args which depend on m3 native properties update correctly', async function (assert) {
      this.owner.register('service:m3-schema', TestSchema);
      this.owner.register(
        'template:components/show-book',
        hbs`<h1 class="title">{{@model.title}}</h1>
    `
      );

      this.store.push({
        data: {
          id: 'urn:li:book:1',
          attributes: {
            title: 'How to Win Friends and Influence People',
          },
        },
      });

      let book = (this.book = this.store.peekRecord('com.example.bookstore.Book', 'urn:li:book:1'));
      await render(hbs`
        {{show-book model=this.book}}
      `);

      let renderedTitle = this.element.querySelector('.title').innerText;
      assert.equal(
        renderedTitle,
        'Title: How to Win Friends and Influence People',
        'component is rendered correctly'
      );
      book.title = 'New title';
      await settled();

      renderedTitle = this.element.querySelector('.title').innerText;
      assert.equal(renderedTitle, 'Title: New title', 'component is rendered correctly');
    });
```

rwjblue (commented):
Chatted a bit with @igorT about this (and my concerns in the prior comment). I think this is fine for now (with the tweak to use `try`/`finally` instead of relying on the nested access to reset), but we should work to refactor away from all of this.

The path forward is _basically_ to migrate to using `@glimmer/tracking/primitives/storage` (which allows us to remove the need for `Ember.get` in the proxy handler's `get` hook, by ensuring that our own tags are property consumed/entangled) via the https://github.com/ember-polyfills/ember-tracked-storage-polyfill. Once we do that then we can make our proxy return `undefined` for both `unknownProperty` and `setUknownProperty`.

Those changes should likely be follow on work after this lands.

tests/unit/model/native-access/tracked-properties-test.js
igorT: Added one manually

rwjblue (approved):

PR #1534, opened 2022-01-07
[bugfix] ensure that errors can be updated when using useUnderlyingErrorsValue

### LINKED ISSUE #1533: [bug] not able to update errors object on payload while using useUnderlyingErrorsValue
https://github.com/hjdivad/ember-m3/blob/45a454f9b38588c9d549bc3a4f41af35ed2226b6/addon/model.js#L737

# Summary

Running `set(model, 'errors', [])` no longer works when using useUnderlyingErrorsValue. It throws the following error.

image

gabrielcsapo:
ensures that we have a setter for errors when using `useUnderylingErrorsValue`.

[fix #1533]

### DIFF
--- a/addon/model.js
+++ b/addon/model.js
@@ -741,6 +741,12 @@ export default class MegamorphicModel extends EmberObject {
     }
     return this._errors;
   }
+
+  set errors(value) {
+    if (this._schema.useUnderlyingErrorsValue(this._modelName)) {
+      return this.setUnknownProperty('errors', value);
+    }
+  }
 }
 
 MegamorphicModel.prototype.store = null;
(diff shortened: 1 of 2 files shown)


addon/model.js
```diff
@@ -741,6 +741,12 @@ export default class MegamorphicModel extends EmberObject {
     }
     return this._errors;
   }
+
+  set errors(value) {
+    if (this._schema.useUnderlyingErrorsValue(this._modelName)) {
+      return this.setUnknownProperty('errors', value);
+    }
```
hjdivad, line 748: 1. can we confirm the semantics of setting `errors` prior to the getter being introduced?  My expectation is that it would update the underlying errors object in place

2. assuming that's right, we should have an else branch here that has those semantics for the case where `useUnderlyingErrorsValue(modelName) === false`

hjdivad (changes requested):

addon/model.js
gabrielcsapo: good point. I added a test and an else branch.

hjdivad (approved):

PR #1568, opened 2022-02-11
Don't deprecate setting symbols

runspired:
in addition to symbols being very useful for stashing debug info, they error when you attempt to cast them to string for the deprecation.


hjdivad (approved):

hjdivad:
Thanks @runspired good catch

PR #1601, opened 2022-04-28
Added more checks when creating nested model

larry-x-yu:
When creating a nested model, currently the framework determines whether to create a nested dirty model based on whether we are creating the model from a server payload. The check is not sufficient as in the case where a user might use model.set('nestedModelAttr', {...nestedModle}). The added check makes sure that this use case is covered when creating the nested model.

Testing done:
All test cases passed in local environment.


hjdivad (changes requested):
1. Couple of style/craft issues inline.

2. The perf task failure is concerning, as the perf task is running successfully in master and in the recently merged branch.  Can we please dig in to see what's going on here?  We don't want to break the perf job nor do we want a detectable perf regression.

hjdivad (approved):

PR #1607, opened 2022-05-09
Move feature-flags from CI to nightly

hjdivad:
We depend on the Ember Data CUSTOM_MODEL_CLASS feature flag, which is
off by default in 3.24.0, and on by default in 3.28.0. We used to have
to test this feature flag against canaries prior to 3.28.0's release,
but no longer.

We no longer depend on any optional feature flags, but still run the job
to ensure that enabling optional features in Ember Data does not
otherwise break ember-m3.

The job will be red right now because we currently are not compatible
with 4.0.0 and also not with canary releases of ember and ember-data.

Co-Authored-By: Larry Yu <laryu@linkedin.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,20 +51,6 @@ jobs:
       - name: test
         run: yarn test:production
 
-  test-feature-flags:
-    name: feature flags
-    needs: [lint]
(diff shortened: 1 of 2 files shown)


larry-x-yu (approved):
LGTM

PR #1693, opened 2022-12-16
Remove implicit injections for Ember 4.0 compat

nlfurniss:
Tests pass locally


rwjblue (commented):
This is essentially a breaking change right (as evidenced by the required test change)?

nlfurniss:
> This is essentially a breaking change right (as evidenced by the required test change)?

Yes correct

Issue #1666, opened 2022-10-24
Meta/Tracking: Ember 4 Compatibility


rwjblue:
- [ ] ~~https://github.com/hjdivad/ember-m3/issues/1650~~
- [x] Change `release-channel` CI job from `allowedToFail`  (essentially reverting https://github.com/hjdivad/ember-m3/commit/fe27037d8b047b68ede96347a015aa0845c37967)
- [x] Change `ember-data-packages-latest` CI job from `allowedToFail` (reverting [5fa11eb](https://github.com/hjdivad/ember-m3/pull/1664/commits/5fa11eb94bdb7a26956d61064af77f90ede268d7) from #1664 )
- [ ] ~~Re-enable test for `watchModelTypes` (revert 8046542 from #1664 )~~ replaced with #1715

scalvert:
Any reason for the Node engine restriction? https://github.com/hjdivad/ember-m3/blob/master/package.json#L121

hjdivad:
@scalvert not really, I think it was over-eager requirements specs + blueprint leftovers. #1702

hjdivad:
@rwjblue separating Ember 4 and Ember Data 4 compatibility.

hjdivad:
will cut a pre-release + close this ticket.

We do need to fix the debug adapter integration to be ember-4 compatible, but the regression is already affecting ember 3.28 with ember-inspector 4.9.1 (i.e. the latest published to the chrome store)

PR #1708, opened 2023-01-27
Bump deps + fix deprecations

hjdivad:
Part of #1666

- Upgrade ember-source to 3.28.11
- Fix LinkTo deprecation
- Fix template deprecations
- Remove bowerrc
- Bump some deps
- jsconfig for LSP configuration
- v6 changelog
- Fix deprecations


hjdivad:
This version of ember-qunit doesn't support 3-24.

That's fine we're going to drop support for 3.x < 3.28 as part of ember 4 support.

PR #1740, opened 2023-03-03
Added a test case to simulate the failing scenario

larry-x-yu:
Draft


addon/record-data.js
```diff
@@ -1025,14 +1027,24 @@ export default class M3RecordData {
...
-      !this.__childRecordDatas[key] ||
-      Array.isArray(this.__childRecordDatas[key])
-    ) {
+    if (!this.__childRecordDatas || Array.isArray(this.__childRecordDatas[key])) {
       return undefined;
     }
+
     let nested = this._childRecordDatas[key];
+    if (!nested) {
+      // Scan mapped keys for potential match
```
hjdivad, line 1036:
```suggestion
      // Scan mapped keys for potential match
      // We have to do this because this._childRecordDatas is keyed on the
      // object property name and not the payload key.  This is just an error.
      // It happens because `resolveValue` asks the user for resolutions
      // [early on](https://github.com/larry-x-yu/ember-m3/blob/7827547ef9afba2fcf03f160ee894b6fb35f56ec/addon/resolve-attribute-util.js#L133)
      // but then (via `resolveSingleValue`) calls
      // [createNestedModel here](https://github.com/larry-x-yu/ember-m3/blob/7827547ef9afba2fcf03f160ee894b6fb35f56ec/addon/resolve-attribute-util.js#L88)
      // with the property the user aksed for -- not the payload key.
      //
      // If we ned to fix this (e.g. if this alias lookup is too slow)
      // `resolveUtil` can tell the record data to track payload key access
      // (and then clear at the end of resolution)  and use this state when
      // determining what key to write to in `createChildRecordData`
```

hjdivad: @larry-x-yu lmk if ☝️ is clear enough

larry-x-yu: Makes complete sense!

PR #1750, opened 2023-06-30
fix: use base model name for type comparison

larry-x-yu:
**Summary**

1. Use base model name in comparison when getting existing child RecordData

**Testing**
`yarn test` passed locally

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -1017,23 +1017,6 @@ export default class M3RecordData {
   }
 
   /**
-   * Returns the derived model name of 'modelName'.
-   *
-   * As ember-m3 always gets the child record data from its _baseRecordData, whose `modelName`
-   * was set to the base model name when being created. This is not always working in certain
-   * cases, as m3 may not have a model definition for the base type. This method would give the
-   * consumers a chance to return the correct model name for this particular record data.
-   *
-   * @param {string} modelName the model name of to find the derived model name for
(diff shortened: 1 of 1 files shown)


addon/record-data.js
hjdivad, line 1: As we discussed OOB, let's instead use `computeBaseModelName` in `_getExistingChildRecordData`.

Although that *should* be redundant it's a helpful safety check.

hjdivad (changes requested):

addon/record-data.js
larry-x-yu: Updated based on our discussion.

addon/record-data.js
```diff
@@ -1055,6 +1055,19 @@ export default class M3RecordData {
...
+     * This is a workaround for now.
+     * As m3 always get the existing child record data from _baseRecordData, thus it
+     * makes sense to use the base model name if it is available; Otherwise just use
+     * the model name - it might be already the base one
+     */
+    let childBaseModelName = (newType && this._schema.computeBaseModelName(newType)) || newType;
+    if (childBaseModelName) {
+      // this is userland API so we have to normalize the name via dasherization
+      newType = dasherize(childBaseModelName);
+    }
```
hjdivad, line 1069: are we sure we only have to do this for `newType` and not the existing type?

hjdivad, line 1069: can we add a test for this?

PR #1754, opened 2024-01-03
Remove Implicit Injections in m3-store

nlfurniss:
This silences the deprecation in apps running Ember 4.

https://deprecations.emberjs.com/v4.x/#toc_implicit-injections

For apps running Ember 4, this does nothing so is safe to remove

### DIFF
--- a/addon/initializers/m3-store.js
+++ b/addon/initializers/m3-store.js
@@ -18,10 +18,6 @@ function initializeDebugAdapter(registry) {
 }
 
 export function initialize(application) {
-  // This should be unnecessary
-  // it is done by the meta package
-  // but it should be done by the store package
-  // https://github.com/emberjs/data/issues/7158
   initializeDebugAdapter(application);
 }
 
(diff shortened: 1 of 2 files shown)


addon/initializers/m3-store.js
```diff
@@ -18,12 +18,6 @@ function initializeDebugAdapter(registry) {
 }
 
 export function initialize(application) {
-  // This should be unnecessary
-  // it is done by the meta package
-  // but it should be done by the store package
-  // https://github.com/emberjs/data/issues/7158
-  application.inject('route', 'store', 'service:store');
-  application.inject('controller', 'store', 'service:store');
```
hjdivad, line 26: This is a breaking change for ember 3.x.  We already have a dep on [ember-compatibility-helpers](https://github.com/ember-cli/ember-compatibility-helpers) so should be able to keep these injections only for ember 3.

hjdivad (approved):

